PR #7, opened 2015-06-01
Fix: Not enough arguments for format string.

VesaKarvonen:
Given code like:

```fsharp
let (%) x = x
```

and pointing at `%` gives an error such as:

```
Debugger entered--Lisp error: (error "Not enough arguments for format string")
  message("val ( % ) : x:'a -> 'a")
  apply(message "val ( % ) : x:'a -> 'a" nil)
  fsharp-ac-message-safely("val ( % ) : x:'a -> 'a")
  fsharp-ac-handle-tooltip("val ( % ) : x:'a -> 'a\n\nFull name: Main.( % )")
  fsharp-ac-filter-output(#<process fsharp-complete> "{\"Kind\":\"tooltip\",\"Data\":\"val ( % ) : x:'a -> 'a\\n\\nFull name: Main.( % )\"}\n")
```

The problem is that the format string was missing and the parameter string was interpreted as the format string.

### DIFF
--- a/fsharp-mode-completion.el
+++ b/fsharp-mode-completion.el
@@ -739,7 +739,7 @@ display a short summary in the minibuffer."
             (if fsharp-ac-use-popup
                 (fsharp-ac/show-popup data)
               (fsharp-ac/show-info-window data)))
-        (fsharp-ac-message-safely (fsharp-doc/format-for-minibuffer data))))))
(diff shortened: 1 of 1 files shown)


VesaKarvonen:
Sigh...  Apparently there are multiple tests that assume that `message` is called incorrectly without a format string.

rneatherway:
Closed by #30, thanks!

PR #16, opened 2015-06-25
Remove BOM from process output

juergenhoetzel:
Even though the BOM from STDOUT is correctly handled by Emacs, there
is also a BOM present in STDERR. And there is no way to seperate STDOUT
and STDERR:

http://www.gnu.org/software/emacs/manual/html_node/elisp/Output-from-Processes.html

Emacs-devel discussion: https://lists.gnu.org/archive/html/emacs-devel/2015-06/msg00556.html

BTW: There might be timing issues, because I did not run into this problem on my faster system.

### DIFF
--- a/fsharp-mode-completion.el
+++ b/fsharp-mode-completion.el
@@ -678,8 +678,10 @@ around to the start of the buffer."
   (with-current-buffer (process-buffer proc)
     (save-excursion
       (goto-char (process-mark proc))
-      (insert-before-markers str)))
-
+      ;; Remove BOM, if present
+      (insert-before-markers (if (string-prefix-p "\ufeff" str)
+				 (substring str 1)
+			       str))))
   (let ((msg (fsharp-ac--get-msg proc)))
     (while msg
       (let ((kind (gethash "Kind" msg))


juergenhoetzel:
Rebased on master

rneatherway:
Ah, sorry I had missed this one. Thanks for the update. Do you know in what situation the BOM will appear on STDERR?

juergenhoetzel:
I could only reproduce this issue on ArchLinux i686 (32 bit) systems. I don't know if there is a relation there.

rneatherway:
Having looked over your test program, do you think it would help if we set `Console.OutputEncoding` to `new Text.UTF8Encoding(false)` in FsAutoComplete? This should disable the BOM output. Could you try that on your system with your test program?

juergenhoetzel:
@rneatherway The workaround doesn't help. The BOM is already written, when the standard streams are created. Digging through the Mono Sources (Console.cs) i found:

		// FIXME: Console should use these encodings when changed
		static Encoding inputEncoding;
		static Encoding outputEncoding;

I still trying to figure out the root cause of this Problem on ArchLinux (i686). On Fedora (i686) no BOM is written.

rneatherway:
OK, it shouldn't really impact performance anyway.

juergenhoetzel:
Just for the record: This is finally fixed in Mono:

https://github.com/mono/mono/pull/2023

rneatherway:
Cool! Good job fixing that. I guess we'll leave it in for the foreseeable future, just in case. I don't think it's doing any harm.

PR #18, opened 2015-08-26
Fix $(ac_archive) target

juergenhoetzel:
Different "curl -o" output name cause rebuilding $(ac_archive) target
each time.

Also define $(bin_d) as order-only-prerequisites to prevent
re-downloading each time $(bin_d) gets a new timestamp (when unziping
the distribution).


rneatherway:
Thanks!

PR #20, opened 2015-08-28
Add C-x C-e as default keybinding for eval

ryrun:
C-x C-e is a default keybinding for eval in elisp and clojure(cider), so why not for fsharp-mode, too?


rneatherway:
That is a good point. Thanks!

PR #28, opened 2015-10-09
Fix fsi silent cd

juergenhoetzel:
I use this setup for the custom:

    (uniquify-buffer-name-style (quote reverse) nil (uniquify))

which leads to 

     #silentCd    @"c:/cygwin/home/lenny/fsharp/test/src/test/Library.fs/test/src/test/fsharp/lenny/home/cygwin/";;
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

     stdin(1,1): error FS2302: Directory    'c:/cygwin/home/lenny/fsharp/test/src/test/Library.fs/test/src/test/fsharp/lenny/home/cygwin/' doesn't exist

because the _buffer-name_ doesn't match the _buffer-file-name_

I also removed a redundant and nested use of _safe-excursion_

### DIFF
--- a/inf-fsharp-mode.el
+++ b/inf-fsharp-mode.el
@@ -168,7 +168,7 @@ Input and output via buffer `*inferior-fsharp*'."
   (save-excursion
     ;; send location to fsi
     (let* (
-          (name (file-truename (buffer-name (current-buffer))))
+          (name (file-truename (buffer-file-name (current-buffer))))
           (dir (file-name-directory name))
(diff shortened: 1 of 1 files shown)


rneatherway:
Thanks again :)

Issue #11, opened 2015-06-09
Emacs Comments (auto-indentation bug)


rneatherway:
_From @xenocons on July 14, 2014 10:59_

Hi,

It seems that after entering a comment using //, on pressing enter emacs-mode indents the cursor as if it were after a =.

Expected behaviour is to place // on the next line automatically after enter is pressed in a line that previously started with a comment.

_Copied from original issue: fsharp/fsharpbinding#658_

rneatherway:
Hi @xenocons, would you consider adding a failing test for this?

rneatherway:
_From @xenocons on September 4, 2014 10:12_

Sorry for the late reply! This is a tricky one, I am not too sure how I would go about writing a failing test for this as it involves control character (newline), I will give it a shot this weekend if i can get some time

rneatherway:
No worries, I'm currently working on tightening up the test infrastructure anyway. I've been thinking about introducing tests for the auto-indentation using [ecukes](https://github.com/ecukes/ecukes)

rneatherway:
Hit enter in the help string in FSharp.AutoComplete/Program.fs. Triggers an error ` (wrong-type-argument characterp t)`

rneatherway:
Closed by #35

PR #34, opened 2015-10-28
Indentation

m00nlight:
Basic automatic indentation using SMIE, there are still some case to polish, like the indentation at the beginning of a function body, which is due to my unfamiliar of SMIE or the limit of SMIE parser ability. I will continue to improve the indentation code.


rneatherway:
Wonderful, thanks for taking the time to take a look at this. I will give it a test run soon.

fsharp-mode-indent-smie.el
```diff
@@ -0,0 +1,95 @@
...
+       	     ("(" exprs ")"))
+       (exprs (exprs ";" exprs)
+	      (expr))
+       (type (type "->" type)
+       	     (type "*" type))
+       (branches (branches "|" branches))
+       (decls (sexp "=" expr)))
+     '((assoc "|"))
+     '((assoc "->") (assoc "*"))
+     '((assoc "let" "fun" "type" "open" "f->"))
```
rneatherway: What does the `f->` here mean?

fsharp-mode-indent-smie.el
```diff
@@ -0,0 +1,95 @@
...
+       (expr ("while" expr "do" expr)
+	     ("if" expr "then" expr "else" expr)
+	     ("for" expr "in" expr "do" expr)
+	     ("for" expr "to" expr "do" expr)
+	     ("try" id "with" branches)
+	     ("match" expr "with" branches)
+	     ("type" expr "=" branches)
+	     ("begin" exprs "end")
+	     ("[" exprs "]")
+	     ("[|" exprs "|]")
```
rneatherway: You can also add `{`, `}` here and further down, which are used for sequence expressions.

fsharp-mode-indent-smie.el
```diff
@@ -0,0 +1,95 @@
...
+
+;; You should have received a copy of the GNU General Public License
+;; along with GNU Emacs; see the file COPYING.  If not, write to
+;; the Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+;; Boston, MA 02110-1301, USA.
+
+(require 'smie)
+
+
+(defcustom fsharp-indent-level 2
```
rneatherway: I know I set it to 2 elsewhere, but I think most people use 4 so we should probably have that as the default.

fsharp-mode-indent-smie.el
```diff
@@ -0,0 +1,95 @@
...
+       (assoc "/" "*" "%")))))
+  )
+
+(defun fsharp-smie-rules (kind token)
+  (pcase (cons kind token)
+    (`(:elem . basic) fsharp-indent-level)
+    (`(:after . "do") fsharp-indent-level)
+    (`(:after . "then") fsharp-indent-level)
+    (`(:after . "else") fsharp-indent-level)
+    (`(:after . "try") fsharp-indent-level)
```
rneatherway: also `finally`

fsharp-mode-indent-smie.el
```diff
@@ -0,0 +1,95 @@
...
+    (`(:after . "with") fsharp-indent-level)
+    (`(:after . "in") 0)
+    (`(:after . ,(or `"[" `"]" `"[|" `"|]")) fsharp-indent-level)
+    (`(,_ . ";") (if (smie-rule-parent-p "begin")
+		     0
+		   (smie-rule-separator kind)))
+    (`(:after . "=") fsharp-inent-level)
```
rneatherway: typo: `fsharp-indent-level`

When would this trigger?

rneatherway:
This is looking really good already in my tests! Thanks a lot. Can you explain a bit more the problem you have with a function definition? If I type `let x =<RET>` then the cursor is indented on the next line, which seems right to me.

I think we should just make this the default because your way is much better, so feel free to disconnect the old stuff and enable this by default in `fsharp-mode.el`.

fsharp-mode-indent-smie.el
rneatherway: Also `<@` and `<@@` for quotations: https://msdn.microsoft.com/en-us/library/dd233212.aspx

m00nlight: Add as you suggest.

fsharp-mode-indent-smie.el
m00nlight: Add `finally` keywords

fsharp-mode-indent-smie.el
m00nlight: Change to default 4 space. But since the `fsharp-mode.el` rely on a lot of function defined in `fsharp-mode-indent.el`. I can not easily remove the relevant part of indentation from the code. Just change the default indentation in the mode also to 4.

fsharp-mode-indent-smie.el
m00nlight: Here I just write `f->` want to distinguish from function "->" with other "->" symbol like in type etc. In that cases, we need to do a manual lexer to distinguish the different type of "->" in context(maybe in the future). I suppose that we can just treat them the same at an early stage.

m00nlight:
@rneatherway I just update the pull request for the indentation. Since the `fsharp-mode.el` rely on a lot of functions defined in `fsharp-mode-indent.el`. I can not find an easy way to remove it. So I just add `fsharp-mode-indent-smie.el` to `fsharp-mode.el` and change the default indent offset in `sharp-mode-indent.el` also to 4.

rneatherway:
For me `<RET>` is bound to `fsharp-newline-and-indent` so I guess it is to do with that. How come you use `C-j`?

It's good this fixes #11.

rneatherway:
The Travis tests are failing, could you fix those please?

m00nlight:
I checked, my `C-j` is bound to `parotid-newline`.  So that maybe the problem.  And the travis error message is that `fsharp-mode-indent-smie-setup` is not defined, but it is actually defined in `fsharp-mode-indent-smie.el`.  So will move the `(require 'fsharp-mode-indent-smie)` to the top in `fsharp-mode.el` solve the problem?

m00nlight:
@rneatherway Fix the travis test problem. :)

rneatherway:
Great! Thanks very much.

PR #46, opened 2015-12-25
fsautocomplete + Mono >= 4.2.1 threadpool bug fix

drvink:


rneatherway:
Thanks a lot for taking the time to submit a PR. Looks like it'll do the trick, but do you think we need to warn the user given that this only affects processes launched by Emacs? I think I'd rather avoid that message every time (I think most users won't bother to disable it).

drvink:
I suppose what we could instead do would be to set the environment variable just before executing fsautocomplete, then unset it so that it doesn't affect other things ran via Emacs--that ought to be transparent to the user, so we wouldn't have to warn about it either.

rneatherway:
Yes, that sounds like the best approach, good idea.

drvink:
Pushed a new version: If `fsharp-ac-using-mono` is non-nil and `MONO_THREADS_PER_CPU` is unset or less than 8, we bind `process-environment` adding `MONO_THREADS_PER_CPU=8` before launching fsautocomplete. This only affects fsautocomplete and leaves the user's environment untouched.

rneatherway:
Looks great. The test failure is unrelated so I'll go ahead and merge. Thanks for your contribution!

PR #49, opened 2015-12-31
Company mode (WIP)

nosami:
Very rough early visibility before I go out for the day :)

Working

- Completions after one char

Not working

- electric-dot
- sometimes returns wrong completions
- electric-backspace
- tests
- doc strings
- makes requests even while the completion popup is open
- currently case sensitive

I'm using it currently with the following settings in my personal config

```elisp
  (eval-after-load 'company
    '(add-to-list 'company-backends 'fsharp-ac/company-backend))
  (setq company-idle-delay 0.03)
  (setq company-minimum-prefix-length 0)
```

I'm not sure if the `add-to-list` should go in the plugin code or not yet.

### DIFF
--- a/fsharp-mode-completion.el
+++ b/fsharp-mode-completion.el
@@ -26,7 +26,6 @@
 (with-no-warnings (require 'cl))
 (require 's)
 (require 'dash)
-(require 'auto-complete)
 (require 'company)
 (require 'json)
 (require 'etags)
@@ -312,7 +311,6 @@ For indirect buffers return the truename of the base buffer."
           (set-process-query-on-exit-flag proc nil)
           (with-current-buffer (process-buffer proc)
             (delete-region (point-min) (point-max)))
-          (add-to-list 'ac-modes 'fsharp-mode)
           proc)
       (error "Failed to launch: '%s'" (s-join " " fsharp-ac-complete-command))
       nil)))
@@ -452,7 +450,6 @@ For indirect buffers return the truename of the base buffer."
   "Regexp for a dotted ident with a raw residue.")
 
 (defun fsharp-ac--residue ()
-  (interactive)
   (let ((result
          (let ((line (buffer-substring-no-properties (line-beginning-position) (point))))
            (- (point)
@@ -488,8 +485,7 @@ The current buffer must be an F# file that exists on disk."
 
      (t
       (and (not (syntax-ppss-context (syntax-ppss)))
-           (eq fsharp-ac-status 'idle)
-           (not ac-completing))))))
+           (eq fsharp-ac-status 'idle))))))
 
 (defvar fsharp-ac-awaiting-tooltip nil)
 
@@ -552,9 +548,6 @@ prevent usage errors being displayed by FSHARP-DOC-MODE."
   (when (and (eq major-mode 'fsharp-mode)
              (fsharp-ac-can-make-request t))
     (fsharp-ac-parse-current-buffer)))
-  ; Perform some emergency fixup if things got out of sync
-  ;; (when (not ac-completing)
-  ;;   (setq fsharp-ac-status 'idle)))
 
 ;;; ----------------------------------------------------------------------------
 ;;; Errors and Overlays
--- a/fsharp-mode-completion.el
+++ b/fsharp-mode-completion.el
@@ -361,11 +361,21 @@ For indirect buffers return the truename of the base buffer."
     (fsharp-ac-make-completion-request)
     (setq company-callback callback)))
 
+(defun fsharp-ac-get-prefix ()
+  (let ((prefix (company-grab-word)))
+    (if (string-match "^[[:space:]]*$" prefix)
+        (when (eq (string (char-before)) ".")
+          prefix)
+          ;; returning nil here passes off completion to any other backends
+          ;; that could be configured
+          )
+    prefix))
+
 (defun fsharp-ac/company-backend (command &optional arg &rest ignored)
     (interactive (list 'interactive))
     (cl-case command
         (interactive (company-begin-backend 'fsharp-ac/company-backend))
-        (prefix (company-grab-word))
+        (prefix (fsharp-ac-get-prefix))
         (ignore-case 't)
         (candidates (cons :async 'fsharp-company-candidates))
         (doc-buffer (company-doc-buffer (fsharp-ac-document arg)))))
--- a/fsharp-mode-completion.el
+++ b/fsharp-mode-completion.el
@@ -526,12 +526,6 @@ prevent usage errors being displayed by FSHARP-DOC-MODE."
   (interactive)
   (pop-tag-mark))
 
-
-(defun fsharp-ac/electric-dot ()
-  (interactive)
-  (insert-char ?. 1)
-  (company-complete))
-
 (defun fsharp-ac/electric-backspace ()
   (interactive)
   (when (eq (char-before) ?.)
--- a/fsharp-mode-completion.el
+++ b/fsharp-mode-completion.el
@@ -366,13 +366,15 @@ For indirect buffers return the truename of the base buffer."
              (eq fsharp-ac-status 'idle))
     (setq company-callback callback)
     (fsharp-ac-make-completion-request)))
+(defun fsharp-ac-add-annotation-prop (s candidate)
+  (propertize s 'annotation (gethash "GlyphChar" candidate)))
 
 (defun fsharp-ac-completion-done ()
   (let ((mapped-completion
     (-map (lambda (candidate)
             (let ((s (gethash "Name" candidate)))
-              (if (fsharp-ac--isNormalId s) (propertize s 'annotation (gethash "GlyphChar" candidate))
-                (s-append "``" (s-prepend "``" s)))))
+              (if (fsharp-ac--isNormalId s) (fsharp-ac-add-annotation-prop s candidate)
+                (s-append "``" (s-prepend "``" (fsharp-ac-add-annotation-prop s candidate))))))
           fsharp-ac-current-candidate)))
     (funcall company-callback (fsharp-company-filter company-prefix mapped-completion))))
 
--- a/fsharp-mode-completion.el
+++ b/fsharp-mode-completion.el
@@ -523,7 +523,7 @@ prevent usage errors being displayed by FSHARP-DOC-MODE."
   (interactive)
   (insert-char ?. 1)
   (unless (company-in-string-or-comment)
-    (company-auto-begin)))
+    (company-complete)))
   ;; (when ac-completing
   ;;   (ac-complete))
 
@@ -786,7 +786,7 @@ around to the start of the buffer."
               data)
         fsharp-ac-status 'acknowledged)
   ;; (fsharp-ac--ac-start :force-init t)
-  (company-auto-begin)
+  (company-complete)
   ;; (ac-update)
   (setq fsharp-ac-status 'idle))
 
--- a/fsharp-mode-completion.el
+++ b/fsharp-mode-completion.el
@@ -368,7 +368,13 @@ For indirect buffers return the truename of the base buffer."
     (fsharp-ac-make-completion-request)))
 
 (defun fsharp-ac-completion-done ()
-  (funcall company-callback (fsharp-company-filter company-prefix fsharp-ac-current-candidate)))
+  (let ((mapped-completion
+    (-map (lambda (candidate)
+            (let ((s (gethash "Name" candidate)))
+              (if (fsharp-ac--isNormalId s) (propertize s 'annotation (gethash "GlyphChar" candidate))
+                (s-append "``" (s-prepend "``" s)))))
+          fsharp-ac-current-candidate)))
+    (funcall company-callback (fsharp-company-filter company-prefix mapped-completion))))
 
 (defun fsharp-ac-get-prefix ()
   (if (char-equal (char-before) ?\s)
@@ -382,6 +388,7 @@ For indirect buffers return the truename of the base buffer."
         (prefix  (fsharp-ac-get-prefix))
         (ignore-case 't)
         (candidates (cons :async 'fsharp-company-candidates))
+        (annotation (get-text-property 0 'annotation arg))
         (doc-buffer (company-doc-buffer (fsharp-ac-document arg)))))
 
 (defconst fsharp-ac--ident
@@ -760,12 +767,7 @@ around to the start of the buffer."
     (setq msg (fsharp-ac--get-msg proc)))))
 
 (defun fsharp-ac-handle-completion (data)
-  (setq fsharp-ac-current-candidate
-        (-map (lambda (candidate)
-                (let ((s (gethash "Name" candidate)))
-                  (if (fsharp-ac--isNormalId s) s
-                    (s-append "``" (s-prepend "``" s)))))
-              data)
+  (setq fsharp-ac-current-candidate data
         fsharp-ac-status 'idle)
   (fsharp-ac-completion-done))
 
--- a/fsharp-mode-completion.el
+++ b/fsharp-mode-completion.el
@@ -375,18 +375,8 @@ For indirect buffers return the truename of the base buffer."
         (interactive (company-begin-backend 'fsharp-ac/company-backend))
         (prefix (company-grab-word))
         (candidates 
-                    (setq fsharp-ac-status 'wait)
-                    (setq fsharp-ac-current-candidate nil)
-                    (clrhash fsharp-ac-current-helptext)
-
-                    (fsharp-ac-parse-current-buffer)
-                    (fsharp-ac-send-pos-request
-                    "completion"
-                    (fsharp-ac--buffer-truename)
-                    (line-number-at-pos)
-                        (+ 1 (current-column)))
-                      (list "foobar" "foobaz" "foobarbaz")
-     )
+          (fsharp-ac-candidate))
+        
     (meta (format "This value is named %s" arg))))
 
 (defconst fsharp-ac--ident
@@ -530,16 +520,19 @@ prevent usage errors being displayed by FSHARP-DOC-MODE."
 
 (defun fsharp-ac/electric-dot ()
   (interactive)
-  (when ac-completing
-    (ac-complete))
+  (insert-char ?. 1)
+  (unless (company-in-string-or-comment)
+    (company-auto-begin)))
+  ;; (when ac-completing
+  ;;   (ac-complete))
 
-  (let ((residue (fsharp-ac--residue))
-        (pt (point)))
-    (when (or (not (eq ?. (char-before)))
-              (not ac-completing))
-      (self-insert-command 1))
-    (unless (eq pt residue)
-      (fsharp-ac/complete-at-point t))))
+  ;; (let ((residue (fsharp-ac--residue))
+  ;;       (pt (point)))
+  ;;   (when (or (not (eq ?. (char-before)))
+  ;;             (not ac-completing))
+  ;;     (self-insert-command 1))
+  ;;   (unless (eq pt residue)
+  ;;     (fsharp-ac/complete-at-point t))))
 
 
 (defun fsharp-ac/electric-backspace ()
@@ -791,8 +784,9 @@ around to the start of the buffer."
                     (s-append "``" (s-prepend "``" s)))))
               data)
         fsharp-ac-status 'acknowledged)
-  (fsharp-ac--ac-start :force-init t)
-  (ac-update)
+  ;; (fsharp-ac--ac-start :force-init t)
+  (company-auto-begin)
+  ;; (ac-update)
   (setq fsharp-ac-status 'idle))
 
 (defun fsharp-ac-handle-doctext (data)
--- a/fsharp-mode-completion.el
+++ b/fsharp-mode-completion.el
@@ -356,15 +356,14 @@ For indirect buffers return the truename of the base buffer."
     (setq fsharp-ac-status 'idle))
 
   (when (and (fsharp-ac-can-make-request 't)
-             (eq fsharp-ac-status 'idle)
-             (not (eq (string (char-before)) " ")))
+             (eq fsharp-ac-status 'idle))
     (setq company-callback callback)
     (fsharp-ac-make-completion-request)))
 
 (defun fsharp-ac-get-prefix ()
   (if (char-equal (char-before) ?\s)
       nil
-    (company-grab-word))
+    (company-grab-word)))
 
 (defun fsharp-ac/company-backend (command &optional arg &rest ignored)
     (interactive (list 'interactive))
--- a/fsharp-mode-completion.el
+++ b/fsharp-mode-completion.el
@@ -367,6 +367,28 @@ For indirect buffers return the truename of the base buffer."
      (setq fsharp-ac-status 'idle)
      fsharp-ac-current-candidate)))
 
+(require 'cl-lib)
+
+(defun fsharp-ac/company-backend (command &optional arg &rest ignored)
+    (interactive (list 'interactive))
+    (cl-case command
+        (interactive (company-begin-backend 'fsharp-ac/company-backend))
+        (prefix (company-grab-word))
+        (candidates 
+                    (setq fsharp-ac-status 'wait)
+                    (setq fsharp-ac-current-candidate nil)
+                    (clrhash fsharp-ac-current-helptext)
+
+                    (fsharp-ac-parse-current-buffer)
+                    (fsharp-ac-send-pos-request
+                    "completion"
+                    (fsharp-ac--buffer-truename)
+                    (line-number-at-pos)
+                        (+ 1 (current-column)))
+                      (list "foobar" "foobaz" "foobarbaz")
+     )
+    (meta (format "This value is named %s" arg))))
+
 (defconst fsharp-ac--ident
   (rx (one-or-more (not (any ".` \t\r\n"))))
   "Regexp for normal identifiers.")
--- a/fsharp-mode-completion.el
+++ b/fsharp-mode-completion.el
@@ -27,6 +27,7 @@
 (require 's)
 (require 'dash)
 (require 'auto-complete)
+(require 'company)
 (require 'json)
 (require 'etags)
 (require 'fsharp-mode-util)
@@ -93,6 +94,9 @@ If set to nil, display in a help buffer instead.")
   "BUFFER's tick counter, when the file was parsed.")
 (defvar fsharp-ac--last-parsed-buffer nil
   "Last parsed BUFFER, so that we reparse if we switch buffers.")
+
+(defvar-local company-prefix nil)
+(defvar-local fsharp-company-callback nil)
 
 (defconst fsharp-ac--log-buf "*fsharp-debug*")
 (defconst fsharp-ac--completion-procname "fsharp-complete")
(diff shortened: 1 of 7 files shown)


rneatherway:
I had a quick go and it seems to only be returning completions for words in the same file (including keywords like `module`). It seems like we might need to a) disable the other backends for F#, and b) use some kind of trick to delay displaying the completions until the async response has come back.

nosami:
There are lots of issues with it right now, sorry. Other backends aren't activated for me right now but I haven't pushed all my code yet.

Really struggling with trying to display the popup from the process filter on the response. I've been debugging company-mode on and off for days :) I can't work out if it's a bug in company-mode or the fsharp plugin. 

Company-complete doesn't show the popup the first time that is invoked from there, but it works on subsequent requests (providing that a response has previously been returned.) 

When it doesn't work, it seems like company's pre-command hook isn't firing https://gist.github.com/nosami/a9876856b6532e8f48ec

nosami:
Also, this seems like I'm going against the grain of how company-mode is supposed to be used. 

Right now, when company-complete is invoked we don't have any completions initially, so I return nil. Later when the completions are returned, company-complete is invoked again from the process filter hook.

company-mode has an async mechanism that uses a callback which should be ideal, but I don't think we can use it because the requests and responses are disconnected.

I was thinking that some sort of event loop would be ideal (long term). You could give each request a sequential ID... and then the server would return the same ID on the response. You would create a dictionary of ids -> callbacks and then when the response comes in, dispatch the callback function and remove it from the dictionary.

Might be quite a bit of work to do that though, and more work than I was wanting to do for this PR, so will keep debugging :)

nosami:
I've managed to get async completions working. It was easier than I thought :) 

It's working fairly well now, but still a few issues to sort out. 
The main one is that I disconnected the electric-dot binding. Invoking company-complete directly forces completions to be synchronous, which can be a little annoying. I still haven't figured out how to make `.` an idle completion.
It's also possible for the completion system to get stuck in the `wait` state. I think this happens when completion is aborted due to timeouts etc. Should be easy to fix.
Another minor issue is that because company-mode is a little more aggressive in completing, the backend server is having to do more work. This can result in it sometimes appear to freeze. I'm not sure if this was ever a problem with auto-complete.

Also, I just discovered that I can accidentally push branches here :O ... now deleted!

nosami:
electric-dot is no longer an issue if you have these settings

```elisp
  (setq company-minimum-prefix-length 0)
  (setq company-auto-complete 't)
```

rneatherway:
> Another minor issue is that because company-mode is a little more aggressive in completing, the backend server is having to do more work (especially as more top level completions are being invoked). This can result in it sometimes appearing to freeze. I'm not sure if this was ever a problem with auto-complete mode.

With `auto-complete` I don't have it configured to send any requests automatically, only if `.` is entered or if the function is called explicitly. After that if you enter more characters the existing results are refined rather than requesting more. I found the performance wasn't great if lots of requests were sent one after the other, which is one reason why I ended up doing it this way (also it matches Visual Studio).

If I type `List.to` how many requests will be made? One per character?

nosami:
`List.to` will trigger when you press `L`, and then again as soon as the previous completion is returned and you continue typing.

I was thinking of maybe trying to disable top level completions. Or maybe trying to discard any previous results when `.` is pressed so that new completions can be fetched.

nosami:
Actually, that's not quite right. When you press `L`, a completion request is triggered, but no more completions are triggered for the word `List` no matter if you wait for the response or not. 

The next trigger point is either the `dot` or some character after that if you type fast.

nosami:
You can configure the minimum prefix length and idle-delay, so it will depend on those settings too.

nosami:
Another possibility could be to enable word completion when at the top level.

nosami:
I've added a gif to give you a feel for how it's working for me. You can also see where the completions choke after typing `List.`

company

Overall, I'm pretty happy with it. I think it's an improvement over auto-complete. Just got a few niggles to work out.

rneatherway:
Thanks for the gif. I agree, that is looking really nice! Do you think it will be easy to add support for displaying documentation?

I don't think I completely understand why it's choking on `List` though. Is that because the request is currently synchronous and the response is pretty slow to a) compute, and b) transmit?

nosami:
Yeah. Docs should be easy enough. I just wanted to make sure it was going to work first :)

It's choking because when I am at the `o` in `List.to` it still hasn't returned the results from `L` which is a top level completion. I know that emacs struggles a little deserialising json but I guess the culprit is the compute time on the server. I need to measure the server side response time and compare.  

I think for now I should just try and abort (ignore) pending requests when pressing the `.` key and it should be ok. I did consider caching top level completions because it should be pretty easy (it's a static list unless an import statement changes or a reference is added) but would rather avoid that if possible.

7sharp9:
Dont forget we do lots of cancellation now in XS now which really helps, plus we only do a parse request on 300ms keyboard latency.

nosami:
Haven't even looked at the fsautocomplete code yet. Not sure if it is cancellable right now without some extra work.

nosami:
Ignoring pending requests on `.` works a treat.

list-completion

nosami:
helptext docs!

rneatherway:
This is looking great! I'll try it out tonight.

nosami:
I'm sure there are plenty of bugs with it. Just let me know what you find.

rneatherway:
It's working nicely here for the main part, but there are still some rough edges to do with automatically requesting completions. For example, if I type `(`, then the top-level completions will be fetched, which is a bit pointless. Can we tune when completions are requested a bit more?

The 0-length prefix is causing other problems. For example if I try to type `let x = ` then I end up with: 

`let XA =AbstractClassAttribute AbstractClassAttribute AbstractClassAttribute AbstractClassAttribute`

because of the toplevel completions.

rneatherway:
@7sharp9 which requests do you cancel, completions, parses, others?

nosami:
Controlling which characters trigger the popup is just a parameter, so will be simple to prevent completion on `(`

Not sure about the `let x = ` issue. This is what I see
letx

nosami:
Just pushed commit to stop completion on `.`

rneatherway:
What should I use to make a gif similar to what you've done so I can demonstrate? I can use either OSX or Linux.

nosami:
I use licecap on OS X. The linux gif was taken using linux in a VM :)

rneatherway:
What's the linux alternative to licecap?

Here's what I was seeing. I think perhaps it is just what happens when you press space, (uses first completion) but I did find it rather surprising.

untitled

nosami:
The most similar thing I've seen to licecap on linux is byzanz but it's just a command line app. You need a helper script to grab the rectangle etc.

nosami:
hmm... I can't reproduce that at all. I wonder if spacemacs is configuring some settings that I don't know about.

nosami:
It looks as if space is selecting the completion for you. it doesn't for me. Do you have any other plugins installed that might be affecting this?

nosami:
letx

rneatherway:
Hmm, I've just tried it on Linux and I don't see the same behaviour. Strange. Space is definitely bound to `self-insert-command` here in any case. However, pressing space does still keep bringing up the list of top-level completions, which it seems to be doing for you as well. I think we should avoid requesting completions if the user hasn't entered any text at all.

nosami:
Yeah, ok. Should be possible. I just need to distinguish between no text entered and dot key has just been entered.

nosami:
With the last commit, pressing space shouldn't trigger completions

rneatherway:
> With the last commit, pressing space shouldn't trigger completions

Do you mean the commit 'Differentiate between '.' being typed and whitespace'? I'm just trying the latest and unless I've made a mistake it's not doing the trick. Space is still requesting completions.

nosami:
When I wrote that comment, I thought that the prefix for both cases was "". I was wrong. I'm still working on it :)

rneatherway:
OK no worries! Is it not possible to have a normal prefix of 3 or so and let `.` trigger a manual completion?

nosami:
It doesn't seem that way. If I wire up the `.` to trigger electric-dot or something similar, then the completion popup becomes synchronous which can be super annoying. 

You can try it now - the function is still available. Try it on a completion that you will know will take some time.

nosami:
I was just thinking that a super simple workaround might be to do `(define-key company-active-map (kbd "<space>") nil)` (untested)

nosami:
I've also got the code set up (but not yet pushed), to remove all the other backends when in F# mode

rneatherway:
> It doesn't seem that way. If I wire up the . to trigger electric-dot or something similar, then the completion popup becomes synchronous which can be super annoying.

I see. For auto-complete I think I solved this by returning nil the first time, then caching the results when they come back and firing the completion command again, this time with completions already available so it returns immediately.

nosami:
That's what I attempted originally. See my 2nd and 3rd comments :)

nosami:
nospace
Ok. It's looking much better this time!

rneatherway:
Yes, definitely at improvement, but still a few glitches:

anim

nosami:
Yeah. I know about that one. I think the list just needs to filtered for
words starting with x in your example.

Haven't quite worked out yet why the filtering doesn't always work.

nosami:
You can try setting `(setq company-require-match 't) which helps with this. But then it prevents you from typing words that aren't in the list.

nosami:
Actually... don't do that. It makes things worse :S Will take a look into the filtering tonight.

nosami:
I was wondering if you'd want to add the GlyphChar to the list too? It can be added to a column on the right.

Company-mode also has 2 forms of docs. 'meta' which is a short string that can be echoed to the minibuffer as well as the full docs that we show now. Possibly useful for something in the future.

rneatherway:
Ah, that's pretty cool, using the GlyphChar sounds great. meta could also be useful, possibly for the type? Not sure.

nosami:
Added filtering. It's much better now.
better

nosami:
screen shot 2016-01-08 at 00 16 33

glyph char

rneatherway:
Just tried it and looking really good now, I haven't spotted any issues. Now we just need to settle on a nice set of defaults and handle the testing.

nosami:
Great! I was thinking that it could work better if fsautocomplete cancelled previous Parse / Completion requests. Still haven't looked at how that works yet, but it if that does need doing, it could be done independently from this.

rneatherway:
Yeah, I'm sure it would improve matters, and we could probably take some hints from how @7sharp9 did it. Tagging all the messages sent with an ID would probably help with this, and for editors to interpret responses as well. Are there any particular problems you're seeing though that makes you think this would be useful?

nosami:
It can get bogged down when working in projects that have lots of references / `open` statements. It generally recovers pretty quickly though.

I also haven't done any testing with type providers - I think anything that results in a long parse time could potentially be an issue.

rneatherway:
Right, so that is the same at the moment. I agree it would definitely be an improvement, but I think it's orthogonal to this piece of work.

nosami:
Will take a look at the tests tonight. Do you still want auto-complete mode putting back? If so, I might be better starting again and re-applying the code from this PR.

rneatherway:
If you think it's easy enough, then let's keep them, but make company the default. If it's going to be a terrible headache then I'm OK with removing it.

nosami:
Thought I'd try and get the tests passing as-is first. I get one failing test (check-completion) that fails with (error "Company not enabled"). 

Any pointers gratefully received! The test passes when ran interactively through ert.

rneatherway:
Well completion works for me when I start it with `make run`, and that *should* be the same environment that the tests are run in (a fresh one in `emacs-fsharp-mode/tmp`). I don't get tooltips though, perhaps we need to depend on the company-quickhelp package as well. These tests have always been a bit fiddly to get working properly :(. When I add `(company-mode)` to the completion test, it outputs:

```
Company backend 'company-bbdb' could not be initialized:
Cannot open load file: no such file or directory, company-bbdb
Company backend 'company-nxml' could not be initialized:
Cannot open load file: no such file or directory, company-nxml
Company backend 'company-css' could not be initialized:
Cannot open load file: no such file or directory, company-css
Company backend 'company-eclim' could not be initialized:
Symbol's function definition is void: company-eclim
Company backend 'company-semantic' could not be initialized:
Cannot open load file: no such file or directory, company-semantic
Company backend 'company-clang' could not be initialized:
Symbol's function definition is void: company-clang
Company backend 'company-xcode' could not be initialized:
Cannot open load file: no such file or directory, company-xcode
Company backend 'company-cmake' could not be initialized:
Symbol's function definition is void: company-cmake
Company backend 'company-capf' could not be initialized:
Symbol's function definition is void: company-capf
Company backend 'company-dabbrev-code' could not be initialized:
Cannot open load file: no such file or directory, company-dabbrev-code
Company backend 'company-gtags' could not be initialized:
Cannot open load file: no such file or directory, company-gtags
Company backend 'company-etags' could not be initialized:
Cannot open load file: no such file or directory, company-etags
Company backend 'company-keywords' could not be initialized:
Cannot open load file: no such file or directory, company-keywords
Company backend 'company-oddmuse' could not be initialized:
Cannot open load file: no such file or directory, company-oddmuse
Company backend 'company-files' could not be initialized:
Cannot open load file: no such file or directory, company-files
Company backend 'company-dabbrev' could not be initialized:
Cannot open load file: no such file or directory, company-dabbrev
Loaded F# project 'Test1.fsproj'
```

which might be helpful?

Also I was trying out the behaviour with backticks and it isn't quite right. The example below shows be trying to complete a backticked identifier from another module, but I can't explicitly request completions, and they aren't offered while typing this. I'm also not sure how to get completions for a backticked identifer from the current module, and completions are offered after I close the double backticks.

```
module N =
  let ``backticked thing`` = 5

let x = N.``bac
```

nosami:
`make run` is a useful tip, thanks - I somehow missed that. I'll spend some more time looking at it tonight and try and get the backtracked completions working.

nosami:
The test is fixed locally and I fixed the byte-compilation errors. However, I now get this when running `make check-declares` :-

```
Checking /usr/share/emacs/24.5/lisp/json.el.gz...OK
Checking /home/jason/src/fsharp/emacs-fsharp-mode/tmp/.emacs.d/elpa/company-20160112.508/yasnippet.el...
Checking /home/jason/src/fsharp/emacs-fsharp-mode/tmp/.emacs.d/elpa/company-20160112.508/yasnippet.el...6 problems found
Checking /home/jason/src/fsharp/emacs-fsharp-mode/tmp/.emacs.d/elpa/company-20160112.508/bbdb.el...
Checking /home/jason/src/fsharp/emacs-fsharp-mode/tmp/.emacs.d/elpa/company-20160112.508/bbdb.el...2 problems found
Checking /home/jason/src/fsharp/emacs-fsharp-mode/tmp/.emacs.d/elpa/company-20160112.508/bbdb-com.el...
Checking /home/jason/src/fsharp/emacs-fsharp-mode/tmp/.emacs.d/elpa/company-20160112.508/bbdb-com.el...2 problems found
Checking /usr/share/emacs/24.5/lisp/emacs-lisp/find-func.el.gz...
uncompressing find-func.el.gz...
uncompressing find-func.el.gz...done
Checking /usr/share/emacs/24.5/lisp/emacs-lisp/find-func.el.gz...OK
Checking /usr/share/emacs/24.5/lisp/emacs-lisp/lisp-mnt.el.gz...
uncompressing lisp-mnt.el.gz...
uncompressing lisp-mnt.el.gz...done
Checking /usr/share/emacs/24.5/lisp/emacs-lisp/lisp-mnt.el.gz...OK
Checking /home/jason/src/fsharp/emacs-fsharp-mode/fsharp-doc.el...
Checking /home/jason/src/fsharp/emacs-fsharp-mode/fsharp-doc.el...OK
Checking /home/jason/src/fsharp/emacs-fsharp-mode/fsharp-mode.el...
Checking /home/jason/src/fsharp/emacs-fsharp-mode/fsharp-mode.el...OK
Warning (check-declare): company-css.el said `web-mode-language-at-pos' was defined in web-mode.el: file not found
Warning (check-declare): company-semantic.el said `semantic-active-p' was defined in semantic.el.gz: function not found
Warning (check-declare): company-semantic.el said `semantic-analyze-possible-completions' was defined in complete.el.gz: function not found
Warning (check-declare): company-yasnippet.el said `yas--table-hash' was defined in yasnippet.el: file not found
Warning (check-declare): company-yasnippet.el said `yas--get-snippet-tables' was defined in yasnippet.el: file not found
Warning (check-declare): company-yasnippet.el said `yas-expand-snippet' was defined in yasnippet.el: file not found
Warning (check-declare): company-yasnippet.el said `yas--template-content' was defined in yasnippet.el: file not found
Warning (check-declare): company-yasnippet.el said `yas--template-expand-env' was defined in yasnippet.el: file not found
Warning (check-declare): company-yasnippet.el said `yas--warning' was defined in yasnippet.el: file not found
Warning (check-declare): company-bbdb.el said `bbdb-dwim-mail' was defined in bbdb-com.el: file not found
Warning (check-declare): company-bbdb.el said `bbdb-search' was defined in bbdb-com.el: file not found
Warning (check-declare): company-bbdb.el said `bbdb-record-get-field' was defined in bbdb.el: file not found
Warning (check-declare): company-bbdb.el said `bbdb-records' was defined in bbdb.el: file not found
Checking `declare-function' statements...13 problems found
Makefile:97: recipe for target 'check-declares' failed
```

Not sure what can be done about this? Looks like some company plugins call functions in some external plugins that we don't care about.

On Travis, there's a completely different issue - `Package 'emacs-24.4' is unavailable`

rneatherway:
Huh, that's annoying. I'll take a look at both of these issues.

rneatherway:
OK, Travis should be working after #53, so you can rebase on that. Regarding the external plugins, I know it's rather tiresome but perhaps it would work if we add the corresponding packages to the `required-packages` list?

nosami:
Yeah... I'm sure that would work. That only gets used by the tests, right?

rneatherway:
Yes, that's right, and in the sandboxed `HOME` under `emacs-fsharp-mode/tmp`, so it should be safe.

nosami:
OK. Will do that, then take a look at the backticked completions.

rneatherway:
OK great. I still have a concern about being able to manually request completions. It seems to me that there are always going to be cases where the user wants to force the completion data to be re-requested, for example at `$` in `let x = (complicated expression).$`, where the type of the bracketed expression may not be available at first. What do you think?

nosami:
I haven't managed to figure out where all the required packages are located. I'm trying to modify the makefile so that only the files in this repo are checked. Does that sound ok?
 
Have you tried to force completion with `C-c C-.` ? I thought that was working.

rneatherway:
Sure, that's the intent of the check anyway. I'll double check `C-c C-.`.

nosami:
There are certainly some bugs with the elisp code regarding backticks... but looks like the backend has some issues too - it's not returning them. I tried the auto-complete plugin too.

screen shot 2016-01-13 at 22 54 08

nosami:
Hi Robin, are you able to verify this? Or am I doing something wrong here?

rneatherway:
Sorry, been a bit busy but I will check it tomorrow.

rneatherway:
The manual completion fetching seems to be working perfectly! My mistake, sorry about that.

What kind of bugs are you seeing with the elisp code and backticks? You are right to say that the completions are not being returned by the backend service, I think this is an issue with FSharp.Compiler.Service.

nosami:
I didn't actually see any bugs - I just assumed it wouldn't be working :) I'm using a company-mode built in function `company-grab-word` [here](https://github.com/fsharp/emacs-fsharp-mode/pull/49/files#diff-ae222c72967dbc27b69ca8868d4878f8R388)

This grabs the current 'prefix' - the word after the `.` or space - but it doesn't take backticks into consideration. I think I'll need to make a new version of this function.

rneatherway:
I wrote some rather annoyingly verbose code to handle this for autocomplete, see here: https://github.com/fsharp/emacs-fsharp-mode/blob/master/fsharp-mode-completion.el#L418

Hopefully that will be of some use (it takes the longest match against normal id or backticked id on the current line). Almost always that ends up being the same as what you describe as the current prefix (after `.` or whitespace).

nosami:
Yeah. That looks like it will be a drop in replacement. I'll give it a go tonight.

rneatherway:
I've been trying it for some general editing and found that it triggers top-level completions on `:` as well. Could we avoid triggering on any symbols?

nosami:
So it does. Yeah, should be easy enough.

nosami:
So now it only triggers on a word-char `(= ?w (char-syntax c))` or a `.` - Not sure about this, but it seems to be doing the right thing.

Also switched the code over to use the fsharp-ac--residue function.

rneatherway:
OK it's looking pretty good. Completions of backticked identifiers still don't seem to work, but as you point at the backend service isn't returning them so we can't help that. It's the same with master. What do you think remains to be done here?

nosami:
I think the backticked completions will work fine as it's using the same function that you were using before. If there are any issues with it, I'll be happy to fix them as they appear.

I think that this is already in a better shape than autocomplete-mode. 

I have only tested this on linux though, so I guess I should test on OSX and Windows too to make sure that the help popup display doesn't have any issues. I don't expect any issues with company-mode itself. I won't be near a machine with Windows until tomorrow evening.

There's also the question of whether or not to support autocomplete-mode, but the farther I got with this, the less incentive there seemed to be to support both.

rneatherway:
I've tested on OSX, and everything is fine there. If you can let me know that you tested on Windows that would be great. I will do a last pass through the code tomorrow. I'm also happy to drop autocomplete-mode support.

nosami:
Great! I'll take a look at it when I get home later.

nosami:
I just tried this on Windows. If the Emacs frame is large enough to accommodate the tooltip, then it works the same as it does on Linux.

image

If the frame is too narrow, then the tooltip covers the completion list.

image

Not sure how much of an issue this is.

rneatherway:
Hmm, well I guess that is a `pos-tip` bug, which was also used by autocomplete-mode, so I wouldn't worry about it.

fsharp-mode-completion.el
```diff
@@ -416,6 +452,7 @@ For indirect buffers return the truename of the base buffer."
   "Regexp for a dotted ident with a raw residue.")
 
 (defun fsharp-ac--residue ()
+  (interactive)
```
rneatherway: Is this necessary?

fsharp-mode-completion.el
```diff
@@ -499,51 +536,25 @@ prevent usage errors being displayed by FSHARP-DOC-MODE."
...
 (defun fsharp-ac--parse-current-file ()
   (when (and (eq major-mode 'fsharp-mode)
              (fsharp-ac-can-make-request t))
-    (fsharp-ac-parse-current-buffer))
+    (fsharp-ac-parse-current-buffer)))
   ; Perform some emergency fixup if things got out of sync
-  (when (not ac-completing)
-    (setq fsharp-ac-status 'idle)))
+  ;; (when (not ac-completing)
+  ;;   (setq fsharp-ac-status 'idle)))
```
rneatherway: If not required, just remove this

fsharp-mode-pkg.el
```diff
@@ -2,8 +2,10 @@
   "fsharp-mode"
   "1.7.2"
   "F# mode for Emacs"
-  '((auto-complete "1.4")
-    (popup      "0.5")
-    (pos-tip    "0.4.5")
-    (s          "1.3.1")
-    (dash       "1.1.0")))
+  '((auto-complete     "1.4")
```
rneatherway: OK we don't need auto-complete anymore, or popup.el?

fsharp-mode.el
```diff
@@ -84,9 +84,9 @@
 
   (define-key fsharp-mode-map "\C-m"      'fsharp-newline-and-indent)
   (define-key fsharp-mode-map "\C-c:"     'fsharp-guess-indent-offset)
-  (define-key fsharp-mode-map [delete]    'fsharp-electric-delete)
-  (define-key fsharp-mode-map [backspace] 'fsharp-electric-backspace)
-  (define-key fsharp-mode-map (kbd ".") 'fsharp-ac/electric-dot)
+  ;; (define-key fsharp-mode-map [delete]    'fsharp-electric-delete)
+  ;; (define-key fsharp-mode-map [backspace] 'fsharp-electric-backspace)
+  ;; (define-key fsharp-mode-map (kbd ".") 'fsharp-ac/electric-dot)
```
rneatherway: If not needed, just delete

fsharp-mode.el
```diff
@@ -214,7 +215,13 @@
...
           compile-command
           syntax-propertize-function
-
+          company-backends
+          company-auto-complete
+          company-auto-complete-chars
+          company-idle-delay
+          company-minimum-prefix-length
+          company-require-match
+          company-tooltip-align-annotations
```
rneatherway: please remove ac-* below

fsharp-mode.el
```diff
@@ -243,13 +250,20 @@
...
   (setq font-lock-defaults '(fsharp-font-lock-keywords))
   (setq syntax-propertize-function 'fsharp--syntax-propertize-function)
+  ; Some reasonable defaults for company mode
+  (setq company-backends (list 'fsharp-ac/company-backend))
+  (setq company-auto-complete 't)
+  (setq company-auto-complete-chars ".")
+  (setq company-idle-delay 0.5)
+  (setq company-minimum-prefix-length 0)
+  (setq company-require-match 'nil)
+  (setq company-tooltip-align-annotations 't)
```
rneatherway: Happy with these?

test/test-common.el
```diff
@@ -79,7 +79,7 @@
 (defconst tests-load-path
   (mapcar 'expand-file-name `(,@load-path "." ".." "./tests")))
 
-(defconst default-dependencies '(popup s dash pos-tip auto-complete noflet faceup))
+(defconst default-dependencies '(popup s dash pos-tip auto-complete noflet faceup company company-quickhelp))
```
rneatherway: please remove autocomplete

fsharp-mode-completion.el
```diff
@@ -27,6 +27,7 @@
 (require 's)
 (require 'dash)
 (require 'auto-complete)
```
rneatherway: rm auto-complete

rneatherway:
All looks sensible, mainly just removing the remaining mentions of auto-complete

fsharp-mode.el
nosami: works for me :)

fsharp-mode-completion.el
nosami: no.. was just for testing

nosami:
I cleaned up the code and also enabled company-quickhelp-mode when it's installed.

No idea why the tests are failing again. `Cannot open load file: ert`. They are working locally.

rneatherway:
The tests pass here as well. Perhaps an issue with emacs master. I don't see it as blocking.

nosami:
Thanks!

mavnn:
This is looks awesome! Many thanks @nosami / @rneatherway !

rneatherway:
And pushed to MELPA. Just wait for the next rebuild and it'll be live!

nosami:
Great!

mavnn:
This is enormously more responsive for me - thanks @nosami !

nosami:
@mavnn glad you like it! I want to spend some time on making it more responsive when I get time.

PR #56, opened 2016-02-24
Use fsautocomplete prefix "StartsWith" filter

juergenhoetzel:
This reduces the "fsautocomplete" response size and thus improves elisp
parsing/filtering speed when using the "completion" command.

```company``` completion is much faster now when using modules with many values.

Maybe we can also get rid of ```fsharp-company-filter```?

BTW: Thanks for the company backend. I never looked back to autocomplete.

### DIFF
--- a/fsharp-mode-completion.el
+++ b/fsharp-mode-completion.el
@@ -94,7 +94,6 @@ If set to nil, display in a help buffer instead.")
 (defvar fsharp-ac--last-parsed-buffer nil
   "Last parsed BUFFER, so that we reparse if we switch buffers.")
 
-(defvar-local company-prefix nil)
 (defvar-local fsharp-company-callback nil)
 
 (defconst fsharp-ac--log-buf "*fsharp-debug*")
--- a/fsharp-mode-completion.el
+++ b/fsharp-mode-completion.el
@@ -351,12 +351,6 @@ For indirect buffers return the truename of the base buffer."
 
 (require 'cl-lib)
 
-(defun fsharp-company-filter (prefix candidates)
-  (if prefix
-    (cl-loop for candidate in candidates
-             when (string-prefix-p prefix candidate 't)
-             collect candidate)))
-
 (defun fsharp-company-candidates (callback)
   (when (eq company-prefix "")
     ;; discard any pending requests as we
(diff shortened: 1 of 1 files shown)


7sharp9:
Nice, theres actually a duplicated column of data that could be removed too if items were just quoted if needed rather than having a complete duplicate of name all the time.

rneatherway:
Yeah it's true, that column for the display name (https://github.com/fsharp/FsAutoComplete/blob/master/FsAutoComplete.Core/CommandResponse.fs#L66) is almost always the same. Perhaps the empty string there could be used to indicate that they are the same.

With this PR, I'm concerned that if you are in the middle of a word like `List.it` and then you ask for completions with filtering then if you use backspace, the completions won't be there for less specific stems than `it`. Does that work?

juergenhoetzel:
> With this PR, I'm concerned that if you are in the middle of a word like List.it and then you ask for completions with filtering then if you use backspace, the completions won't be there for less specific stems than it. Does that work?

Works for me. A new completion request is made for ```List.i```. This request was also made without my commit. If I understood correctly ```company```only caches candidates as completion progresses.

nosami:
Yeah, that's right. `company-mode` will make a new request whenever the popup isn't currently visible.

This change looks good to me. You are correct about the `fsharp-company-filter` function. It should be removed with this change.

rneatherway:
Right, I see. That was different with the old autocomplete code. Please go ahead and remove the function as well.

nosami:
Just been testing this. 

@rneatherway's point is valid
emacs-company

Completion isn't re-triggered on backspace. It is triggered on the next keystroke though, so I'm not sure how much of a problem this is.

nosami:
Just tested the code without this change, and it has the same behaviour.

juergenhoetzel:
> Just tested the code without this change, and it has the same behaviour.

Removing ```fsharp-company-filter``` also fixed the issue. The root cause was the use of ```company-prefix``` in this filtering function. When using backspace the value of ```company-prefix```  was ```nil```.

I don't know if this *var* should be used by backends anyway. It's not documented.

nosami:
No need for `company-prefix` with your change anyway :) Just tested, looks good to me, and the tests are passing for me locally.

rneatherway:
Hmm, some trouble with Travis. Thanks for testing, let's go ahead.

PR #60, opened 2016-03-07
Unbreak company support on non-graphic displays

drvink:
`company-quickhelp` uses `pos-tip`, which doesn't support terminals.


drvink:
FYI: the current state of things (i.e. use of `company-quickhelp`) is actually a feature regression, since the previous stuff (which used `popup`) supported terminals. :(

rneatherway:
I see, sorry to hear that. What should we do then? Company can show the help in another buffer, which I think is its default (without `company-quickhelp`). Does that work for you?

drvink:
That seems fine to me. I just figured it was worth bringing up--overall, the switch to `company` is a welcome change!

nosami:
Pressing F1 on a completion will show the help text in a split buffer.

rneatherway:
OK, thanks for the PR!

PR #61, opened 2016-03-10
Fix async timeout no project

juergenhoetzel:
Fixes completion errors when the current file is not part of a Project:
`
(error "Company: backend fsharp-ac/company-backend async timeout with args (candidates Sys)")
`


rneatherway:
Nice job, thanks!

PR #69, opened 2016-04-07
prefer exact case sort

nosami:
prefer-exact-case

This fixes a little annoyance I have when using company-mode. It sorts based on the case of the completion.

It doesn't always seem to work, but still seems like an improvement over the current code.


rneatherway:
Looks good, thanks.

PR #70, opened 2016-04-10
Forge + Version mention

sideshowcoder:
- Mention forge for project management as it is probably more usable than
  editing .fsproj files by hand
- Mention emacs mode works with F# 3.0 or higher


rneatherway:
Thanks, good points.

PR #71, opened 2016-04-20
fontify the doc string

nosami:
screen shot 2016-04-20 at 23 55 35

### DIFF
--- a/fsharp-doc.el
+++ b/fsharp-doc.el
@@ -38,6 +38,7 @@
 (with-no-warnings (require 'cl))
 (require 'fsharp-mode-completion)
 
+(declare-function fsharp-mode "fsharp-mode.el")
 (defvar fsharp-doc-idle-delay 0.5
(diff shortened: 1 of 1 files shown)


rneatherway:
Very nice. I realise now this should have been updated to use the `typesig` command, but that's a separate issue.

rneatherway:
There is a compilation error:

    fsharp-doc.el:153:1:Error: the function `fsharp-mode' is not known to be defined.

I think you could fix it with the appropriate `declare-function` as it makes the dependencies cyclic.

nosami:
Ah cool... I spent 10 minutes on the code and then 2 hours trying and failing to work out how to fix the test :)

rneatherway:
Sorry to hear that! Feel free to ping me about such problems.

nosami:
Should be good now.

rneatherway:
Great!

PR #75, opened 2016-04-27
Do not change current buffer when starting FSI

### LINKED ISSUE #72: arrayp, nil on first load of a file to interactive
### Description

On loading a buffer to interactive, using the load-buffer command, or C-x h then C-c C-r it fails, with Wrong type argument arrayp, nil.

### Repro steps

Seleect a a region in the buffer and send it to the fsi. Easiest if done with the whole file C-x h -> C-c C-r.  

### Expected behavior

When I load a buffer to interactive I expect it to load first time. 

### Actual behavior

I get the following error, 

screen shot 2016-04-26 at 20 15 38

### Known workarounds

Can just send the region a second time and it works.. 

### Related information 

* MAC OSX 10.11.3
* MELPA - Stable
*  GNU Emacs 24.5.1
* Mono - 4.2.2.30

rneatherway:
Fixes #72 

@colinbull this should address your problem, could you verify please?

### DIFF
--- a/inf-fsharp-mode.el
+++ b/inf-fsharp-mode.el
@@ -113,10 +113,10 @@ be sent from another buffer in fsharp mode.
     (setq inferior-fsharp-program cmd)
     (let ((cmdlist (inferior-fsharp-args-to-list cmd))
           (process-connection-type nil))
-      (set-buffer (apply (function make-comint)
-                         inferior-fsharp-buffer-subname
-                         (car cmdlist) nil (cdr cmdlist)))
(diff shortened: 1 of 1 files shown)


colinbull:
Yep perfect thanks, I was gonna have a look at this tomorrow, saved me the trouble.

nosami:
I just tested this. Works for me. I was seeing the same issue previously.

rneatherway:
Thanks guys, will release soon.

Issue #74, opened 2016-04-27
Type annotation syntax highlighting not working properly


Holtiton:
### Description

Syntax highlighting for type annotations (at least in record types) seems to work incorrectly. The regex does not terminate matching until some clause I have not figured out. You seem to be able to temporarily and randomly to fix this by re-indenting the affected lines.
### Repro steps
1. Create test .fsx file with following

```
type RecordTest1 = { something: int
                     another: string }

type RecordTest2 = { something :int; another :string }

type RecordTest3 = { something : float; another: float; third :float; }

type RecordTest4 = {
    something: int
    another: string }

type RecordTest5 =
    { something: int
      another: string }

type RecordTest6 =
    {
        something: int
        another: string
        third: Option<int>
    }

type RecordTest7 =
    {
    something: int
    another: string
    third: int option
    }
```
1. Start M-x re-builder and paste in the following  (from [fshapr-mode-font.el](https://github.com/fsharp/emacs-fsharp-mode/blob/master/fsharp-mode-font.el#L173) )

```
"[^:]:\\s-*\\(\\<[A-Za-z_'][^,)=<-]*\\)\\s-*\\(<[^>]*>\\)?"
```
### Expected behavior

This is what I got tinkering with the re-builder that seemed reasonable.
![Proper matching](https://cloud.githubusercontent.com/assets/1980127/14844648/869fb48e-0c61-11e6-9e23-feb7700ee5de.PNG)
### Actual behavior

Regex matching does not end.
![Current](https://cloud.githubusercontent.com/assets/1980127/14844653/93a6e012-0c61-11e6-93f6-5f46855a6a21.PNG)
### Known workarounds

This regex seemed to provide better results for me but I have not tested extensively as I just started using F# so I might be missing some cases.

```
"[^:]:\\s-*\\(\\<[A-Za-z0-9_']*[^ ;\n,)=<-]\\)\\(<[^>]*>\\)?"
```
### Related information
- OS version: Windows 10
- Emacs version: GNU Emacs 25.0.50.1 (x86_64-w64-mingw32) of 2015-11-10
- .NET Runtime version: 4.6.1

rneatherway:
Closed by #76

PR #76, opened 2016-04-27
Record type highlighting

rneatherway:
Ref #74

@Holtiton thanks for the good investigation. You can see here how added a new test and tried your fix, and it seems better but not quite perfect. If you install the `faceup` package, `eval-buffer` the test file and `M-x ert-run-tests-interactively` you can try it.

You'll see a difference of
```
      ((on-line 273
                ("        «k:let» «k:rec» «f:f» («v:searchPaths»: «t:SearchPath list») («v:ptr»: «t:nativeptr»«n:<searchpath_t>») =")
                ("        «k:let» «k:rec» «f:f» («v:searchPaths»: «t:SearchPath» «v:list») («v:ptr»: «t:nativeptr»«n:<searchpath_t>») =")))))
```

Which is indicating that in a case like:

```
type RecordTest7 =
    {
    something: int
    another: string
    third: int option
    }
```

The `option` in `int option` is no longer highlighted. Any thoughts on how we could tweak the regex further?

By the way, the syntax highlighting code is very old and a bit of a mess (before I was maintainer). I think it is inherited from OCaml. I would really like to redo it properly. At least this test framework could help us to keep things under control.

### DIFF
--- a/fsharp-mode-font.el
+++ b/fsharp-mode-font.el
@@ -170,7 +170,7 @@
   `(,fsharp-member-function-regexp 1 font-lock-function-name-face)
   `(,fsharp-overload-operator-regexp 1 font-lock-function-name-face)
   ;; `(,fsharp-constructor-regexp 1 font-lock-function-name-face)
-  `("[^:]:\\s-*\\(\\<[A-Za-z_'][^,)=<-]*\\)\\s-*\\(<[^>]*>\\)?"
+  `("[^:]:\\s-*\\(\\<[A-Za-z0-9_']*[^ ;\n,)=<-]\\)\\(<[^>]*>\\)?"
     (1 font-lock-type-face)             ; type annotations
     ;; HACK: font-lock-negation-char-face is usually the same as
     ;; 'default'. use this to prevent generic type arguments from
--- a/fsharp-mode-font.el
+++ b/fsharp-mode-font.el
@@ -170,7 +170,7 @@
   `(,fsharp-member-function-regexp 1 font-lock-function-name-face)
   `(,fsharp-overload-operator-regexp 1 font-lock-function-name-face)
   ;; `(,fsharp-constructor-regexp 1 font-lock-function-name-face)
-  `("[^:]:\\s-*\\(\\<[A-Za-z0-9_']*[^ ;\n,)=<-]\\)\\(<[^>]*>\\)?"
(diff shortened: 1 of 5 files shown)


nosami:
I made a bunch of syntax highlighting tests here https://github.com/fsharp/xamarin-monodevelop-fsharp-addin/blob/master/MonoDevelop.FSharp.Tests/SyntaxHighlighting.fs They could be adapted for emacs.

Holtiton:
This should highlight the words after type annotations
```
"[^:]:\\s-*\\(\\<[A-Za-z0-9_']*[^ ;\n,)=<-]\\)\\([ ]\\w*\\)?\\(<[^>]*>\\)?"
```
So either that or if F# only has syntactic sugar for a few types we could list them separately and should not get any surprises. This might break in some cases but it seemed to work.

Also as this is a new group, a new match is also needed. Don't know how this affects the HACK part in the match. So the complete section would be:
``` 
 `("[^:]:\\s-*\\(\\<[A-Za-z0-9_']*[^ ;\n,)=<-]\\)\\([ ]\\w*\\)?\\(<[^>]*>\\)?"
    (1 font-lock-type-face)             ; type annotations
    (2 font-lock-type-face)             ; type annotations
    ;; HACK: font-lock-negation-char-face is usually the same as
    ;; 'default'. use this to prevent generic type arguments from
    ;; being rendered in variable face
    (3 font-lock-negation-char-face nil t))
```

rneatherway:
I was a bit surprised to find that you can do this with any generic type, not just `list` and other types from OCaml:

```
type Holder<'A> =
    | Thing of 'A

let f (x: Holder<int>) = x
let g (x: int Holder) = x
```

Your suggestion doesn't work for me, it doesn't highlight the normal single word types. I'm trying:

    "[^:]:\\s-*\\(\\<[A-Za-z_'][^\n;,)=<-]*\\)\\s-*\\(<[^>]*>\\)?"

rneatherway:
Ignore that last one, let me know what you think of the latest version of this PR.

Holtiton:
Seems good enough. I'm happy for now with it not matching on multiple lines as that was not pretty with my theme. 

Thanks for the fix and I'll look into at least organizing the regexes soon.

rneatherway:
Cool, that would be great! The `faceup` module is very useful for testing this, let me know if you need help with it. @nosami's tests could also be very useful.

Issue #80, opened 2016-05-11
check-project-files test fails on windows


fradav:
### Description

check-project-files test fails

### Repro steps

run the integration test

### Expected behavior

should pass

### Actual behavior

fails because of hardcoded `"/"` directory separator in the string-match, when windows encompass `"\\"`. 

### Known workarounds

replace with regexp `"\(\\\\\|/\)"`

rneatherway:
Ah, I haven't run the integration tests on Windows for a while, because they don't pass on Appveyor due to an Emacs bug. If you have the workaround available, please do submit a PR.

rneatherway:
Fixed by #81

PR #81, opened 2016-05-11
Fix check-project-files on Windows

juergenhoetzel:
Need to convert expected filename to an OS filename. Also use
`regexp-quote' because Windows backslash need to be escaped in regular
expressions. Refs #80.


rneatherway:
I assume you tested it on Windows?

juergenhoetzel:
Yes. Both: Gnu Linux and Windows

rneatherway:
Great, thanks

PR #83, opened 2016-05-19
Fix completion of type annotated symbols

juergenhoetzel:
Using type annotation without white-space triggers these errors.

First case of failure (wrong replacement):
```
type Dummy = Foo | Bar
let x:D
       ^ complete
->
let Dummy

```
the identifier is also replaced because `fsharp-ac-get-prefix` also matches the text before the colon.

Second case of failure (args-out-of-range error):

```
let dummy:D
           ^complete
->
Debugger entered--Lisp error: (args-out-of-range #("DBNull" 0 6 (annotation "C")) 0 7)
  substring(#("DBNull" 0 6 (annotation "C")) 0 7)

```

this is because the prefix i"dummy:D" is longer than the candidates returned by _fsautocomplete.exe_

### DIFF
--- a/fsharp-mode-completion.el
+++ b/fsharp-mode-completion.el
@@ -489,7 +489,7 @@ If HOST is nil, check process on local system."
            (or (regexp ,fsharp-ac--ident)
                (regexp ,fsharp-ac--rawIdent))
            "."))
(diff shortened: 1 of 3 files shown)


rneatherway:
Nice, thanks!

PR #84, opened 2016-05-21
Don't clobber company-backends

nosami:
This is a little more complicated than I expected because company-backends can be a list of lists.

I'm removing `company-dabbrev`, `company-dabbrev-code` and `company-keywords` from `company-backends` when in fsharp-mode instead of just allowing the fsharp backend.

Before (stock spacemacs backends)
```elisp
(fsharp-ac/company-backend company-files company-oddmuse (company-dabbrev-code company-gtags company-etags company-keywords) company-capf company-cmake company-xcode company-clang company-semantic company-eclim company-css company-nxml company-bbdb)
```

After

```elisp
(fsharp-ac/company-backend company-bbdb company-nxml company-css company-eclim company-semantic company-clang company-xcode company-cmake company-capf (company-gtags company-etags) company-oddmuse company-files)
```

/cc @juergenhoetzel

### DIFF
--- a/fsharp-mode.el
+++ b/fsharp-mode.el
@@ -248,14 +248,16 @@
   (setq font-lock-defaults '(fsharp-font-lock-keywords))
   (setq syntax-propertize-function 'fsharp--syntax-propertize-function)
   ; Some reasonable defaults for company mode
-  (setq company-backends (list 'fsharp-ac/company-backend))
   (setq company-auto-complete 't)
   (setq company-auto-complete-chars ".")
   (setq company-idle-delay 0.03)
   (setq company-minimum-prefix-length 0)
   (setq company-require-match 'nil)
   (setq company-tooltip-align-annotations 't)
-  
+  (setq company-backends
+        (let ((backends-to-remove '(company-dabbrev company-dabbrev-code company-keywords)))
+              (remove-all #'(lambda (backend) (member backend backends-to-remove)) company-backends)))
+  (add-to-list 'company-backends 'fsharp-ac/company-backend)
 
   ;; Error navigation
   (setq next-error-function 'fsharp-ac/next-error)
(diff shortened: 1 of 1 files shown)


juergenhoetzel:
Works for me :+1: 
But the _global_ `company-backends` is still _clobbered_. I.e. The backends are also disabled in other modes.

nosami:
Oh strange. I tested that too.

nosami:
@juergenhoetzel I just tested this and it's working fine here. I went to an F# buffer, saw the reduced set of backends, then switched to an xml buffer and saw the full set.

`company-backends` is set to a buffer-local variable [here](https://github.com/nosami/emacs-fsharp-mode/blob/56b00a0c05460d2387dda9627cd6f40c065b2575/fsharp-mode.el#L215), so not sure how this could happen

juergenhoetzel:
> company-backends is set to a buffer-local variable here, so not sure how this could happen

Sry. You are right. I _clobbered_ too much in my running emacs session :disappointed:

nosami:
No problem. Just glad it's working!

juergenhoetzel:
Dabbrev could still be useful in F# mode:

It seems the root cause of this issue is that the completion is passed to the other backends even if it doesn't make sense. For example after whitespace:

```
let 
    ^ point here 
    
```
According to the [docs](https://github.com/company-mode/company-mode/blob/master/company.el#L333) (if I understand correctly) `'stop` should be returned in this case.

I changed the prefix handler to only pass the completion after whitespace to other backends, if and only if  `(point)` is inside a string or comment.

What do you think?

nosami:
So you want to be able to complete words when inside strings? Seems reasonable.

nosami:
Let me try your commit. If that works ok, then I guess the only thing to do is close this PR and just add the F# backend to the list.

nosami:
OK. Closing this. I like your solution better :)

juergenhoetzel:
> So you want to be able to complete words when inside strings? Seems reasonable.

Yes, but there are still issues. For example completion is still passed to other backends after string literals. For example:

```
let x = "X"<-point here
```
I need to check this out

nosami:
That's working fine here.

juergenhoetzel:
> That's working fine here.

On my system `company-dabbrev-code` is invoked and returns candidates.
I will create a new pull request and remove the check for whitespace to further restrict the pass to other backends.

PR #86, opened 2016-05-31
Dont modify company-transformers

juergenhoetzel:
This variable is intended to be customized by users. 

Instead sort candidates in `fsharp-mode-completion-done` and return `t` for company
command `sorted`.

I hit this issue when editing a `non-F#` mode file and `company-complete`  was triggered:

          
```
Debugger entered--Lisp error: (args-out-of-range "local" 0 6)
  substring("local" 0 6)
  (equal prefix (substring candidate 0 plen))
  (if (equal prefix (substring candidate 0 plen)) (setq sameprefix (cons candidate sameprefix)) (setq others (cons candidate others)))
  (if (equal prefix candidate) (setq exact (cons candidate exact)) (if (equal prefix (substring candidate 0 plen)) (setq sameprefix (cons candidate sameprefix)) (setq others (cons candidate others))))
  (lambda (candidate) (if (equal prefix candidate) (setq exact (cons candidate exact)) (if (equal prefix (substring candidate 0 plen)) (setq sameprefix (cons candidate sameprefix)) (setq others (cons candidate others)))))("local")
  mapc((lambda (candidate) (if (equal prefix candidate) (setq exact (cons candidate exact)) (if (equal prefix (substring candidate 0 plen)) (setq sameprefix (cons candidate sameprefix)) (setq others (cons candidate others))))) ("local"))
  (let* ((exact nil) (sameprefix nil) (others nil) (prefix (fsharp-ac-get-prefix)) (plen (length prefix))) (mapc (function (lambda (candidate) (if (equal prefix candidate) (setq exact (cons candidate exact)) (if (equal prefix (substring candidate 0 plen)) (setq sameprefix (cons candidate sameprefix)) (setq others (cons candidate others)))))) candidates) (append exact sameprefix others))
  fsharp-company-sort-preferring-exact-or-same-case(("local"))
  company--transform-candidates(("local"))
  company--postprocess-candidates(("local"))
```

`fsharp-company-sort-preferring-exact-or-same-case` was called and triggered an error because `fsharp-ac-get-prefix` returned nil.


juergenhoetzel:
@nosami Could you checked if kept the sorting properties (exact first -> same prefix -> rest) you had in mind:

[New Sorting](https://github.com/fsharp/emacs-fsharp-mode/pull/86/commits/ce81f95175fe9d0a3b64f6e6c96f9813524939a2#diff-ae222c72967dbc27b69ca8868d4878f8R412)

Thanks

nosami:
I use spacemacs, and by default it sets `company-transformers` to `(spacemacs//company-transformer-cancel company-sort-by-occurrence)`

If I remove `company-sort-by-occurrence` then all is fine, but if I don't then the list is sorted again after it hits your new code. I can't decide if this is a bug in the config or not.

juergenhoetzel:
> If I remove company-sort-by-occurrence then all is fine, but if I don't then the list is sorted again after it hits your new code. I can't decide if this is a bug in the config or not.

I suspect: `fsharp-company-sort-preferring-exact-or-same-case`  was the last item in `company-transformers` so it had the highest priority in sorting?
But I wonder what's the use case of having multiple sorting transformers.

nosami:
You're right, there shouldn't be more than one sorting transformer in the list. 

We probably should remove `company-sort-by-occurrence` if it exists? It's breaking your new code.

juergenhoetzel:
I use GNU Emacs, which has no default transformers setup. So it didn't broke my setup. 

But I thought about it again:

It is possible generalize fsharp-company-sort-preferring-exact-or-same-case to something like this (untested):

```
(defun company-sort-preferring-exact-or-same-case (candidates)
  (->> (--group-by (cond
		    ((equal it company-prefix) 'equal)
		    ((s-starts-with-p company-prefix it) 'starts-with)
		    (t 'rest)) candidates)
       (funcall (lambda (alist) (append (cdr (assoc 'equal alist))
					(cdr (assoc 'starts-with alist))
					(cdr (assoc 'rest alist)))))))
```

This could be used with any company-backend. Maybe we should just remove the transforming code from  `fsharp-mode`?

nosami:
You'd need to distribute that sort code somehow though... maybe via PR to company-mode. It might need some modification first as it is assuming that the list is already pre-sorted.

I'd be happy with your PR as-is. I can always submit a fix to the fsharp-layer in spacemacs.

juergenhoetzel:
> You'd need to distribute that sort code somehow though... maybe via PR to company-mode. It might need some modification first as it is assuming that the list is already pre-sorted.
> 
> I'd be happy with your PR as-is. I can always submit a fix to the fsharp-layer in spacemacs.

I see. :+1. There should be a good default for the order of candidates.

nosami:
Just a thought, but couldn't `company-transformers` be added to the list of buffer local variables?

juergenhoetzel:
> Just a thought, but couldn't company-transformers be added to the list of buffer local variables?

This was also my first thought :smile: But I realized,that this would just be another workaround which may irritate users.

IMHO custom variables like `company-transformers` should never modified  programmatically by 3rd party code. Making it localy may even lead to more irritation, because the customization in a running Emacs using `customize-variable` has no effect.

Maybe @dgutov can bring clarification?

dgutov:
> company-transformers should never modified programmatically

Indeed. Maybe not "never", but you should have a really good reason to do that.

Doesn't `sorted` work for your use case?

juergenhoetzel:
> company-transformers should never modified programmatically
> Indeed. Maybe not "never", but you should have a really good reason to do that.
> 
> Doesn't sorted work for your use case?

`sorted` works fine. The backend already returns sorted candidates.

This pull request is about the code which @nosami introduced to  do some reordering [prefering exact-match and same case](https://github.com/fsharp/emacs-fsharp-mode/pull/86#issuecomment-222791187),

dgutov:
I see. For distribution, either way is fine:

* Distribute it with `fsharp-mode` and mention it in the documentation.
* Or make a pull request to add it to company.

In both cases the users should add it to `company-transformers` themselves.

juergenhoetzel:
The real reason why we need this transformer is the fact that [FsAutoComplete](https://github.com/fsharp/FsAutoComplete) returns
case insensitive completions (unlike the backends of company-merlin and company-ghc).

I personally like this behavior, because i often use the wrong case in prefix.

So I propose a [new transformer](https://github.com/company-mode/company-mode/pull/531) which might be useful for other
backends returning case-insensitive candidates (for example
`company-dabbrev` if `company-dabbrev-downcase` is turned off).

Actually the part of to original code that checks for exact match is
unnecessary, because the exact match is also the first case
sensitive prefix match.

rneatherway:
Hey, what's the status of this PR? Are we waiting for the one over at company?

juergenhoetzel:
> Hey, what's the status of this PR? Are we waiting for the one over at company?

Yes. I would prefer waiting for the functionality of the [new transformer](https://github.com/company-mode/company-mode/pull/531) merged into [company-mode](https://github.com/company-mode/company-mode)

So i can clean up/rewrite this pull request.

rneatherway:
Cool, sounds good. No hurry.

juergenhoetzel:
The functionality of the [new generic transformer](https://github.com/company-mode/company-mode/pull/531) was merged into [company-mode](https://github.com/company-mode/company-mode)

So i rebased/rewrote this branch.

I also added a short note about the [transformer setup](https://github.com/fsharp/emacs-fsharp-mode/pull/86/commits/2602ddf1392e87f01dc40bef99918b33b8ca6ca7) in the _README_

rneatherway:
Great stuff!

PR #89, opened 2016-07-10
Add tab size manipulation note in README

kuznero:
Simple but somewhat important for those who are not using default 4 spaces per tab.


rneatherway:
Thanks :)

Issue #92, opened 2016-09-21
Evaluate current toplevel phrase on simple function calls doesn't work


ryrun:
### Description

Evaluate current top level phrase on simple function calls doesn't work.

### Repro steps

- Create a fsharp script file, test.fsx:

```fsharp
let test() =
    printfn "Test"

test()

[<EntryPoint>]
let main argv =
    0
```
- Evaluate the function definition
- Move cursor to the end of "test()", press C-M-x (Evaluate current toplevel phrase)

### Expected behavior

No error.

### Actual behavior

Error FS0010: Unerwartete(s/r) Symbol ";;". in Interaktion (Not expected symbol)

### Known workarounds

Mark the whole line and use C-c C-r. (Evaluate region)

### Related information 

* Operating system: Windows 10
* Emacs version: 25.1.1
* Microsoft (R) F# Interactive, Version 14.0.23413.0
* fsharp-mode - 20160719.315

juergenhoetzel:
> Move cursor to the end of "test()", press C-M-x (Evaluate current toplevel phrase)

Does it work, if you add a newline after ```test()```?

ryrun:
When the newline is missing, then i get no ouput. So a newline is needed here.
But i've see that some lines are missing in my test code above. I changed that.

juergenhoetzel:
I can confirm @rneatherway : `fsharp-end-of-block` doesn't seem to handle F# attributes like  `[<EntryPoint>]` correctly and moves beyond the end of the Attribute line.

rneatherway:
Fixed by #93

PR #94, opened 2016-09-22
another fix for fsharp-end-of-block

ryrun:
I removed "\" from regex, it seems escaping is not needed here and doesn't escape anything.

Works on the last line in a file now, too. See comment https://github.com/fsharp/emacs-fsharp-mode/issues/92#issuecomment-248735553


rneatherway:
Looks good, thanks.

PR #99, opened 2016-10-01
Enable IMenu Support

### LINKED ISSUE #98: `imenu` definitions are created but never assigned
### Description

`fsharp-mode-font` defines `fsharp-imenu-generic-expression`, but never actually wires it up to `imenu`. Thus, any attempt to use `imenu` errors. 

### Repro steps

Call `M-x imenu`. Instead of interacting with imenu, one will either receive an error (Emacs 24), or be asked for a variable by which to index (Emacs 25). 

### Expected behavior

Er, imenu should be there.

Gastove:
Wires up existing imenu variables so they'll actually "work"; closes #98. All tests pass. Bumps version to 1.9.2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.9.2 (2016-09-30)
+Features
+  - #98 Enable imenu support
+
 ## 1.9.1 (2016-07-19)
 
 Features:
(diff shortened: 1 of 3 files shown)


rneatherway:
Great! Thanks.

Issue #102, opened 2016-10-17
Stop completions inside of comments


nicholas-peterson:
### Description

Often I'll be writing a comment using '///','//', or '(* .. *)' and want to enter a carriage return or tab. If company mode suggests a completion right as I'm doing this, it enters the completion rather than the desired keystroke. It strikes me as rare that someone would need the completion inside a comment, and wondered if it would help anyone else to simply not trigger completions inside comments?

### Repro steps

+ Open a .fsx file with a project and start entering a comment. 
+ Then place a space, or single character that would start a completion as the end of the line being entered. For instance ' ' or 'a'. This starts the completion menu from company mode.
+ Then hit enter.

### Expected behavior

Instead, the completion menu would never trigger because we're inside a comment block.

### Actual behavior

The completion menu triggers, suggests a completion, and that keyword is entered rather than a carriage return.

### Known workarounds

Cancelling company mode completion using C-g seems to work, just taxing to perpetually do on multi-line comments.

### Related information 

* Windows 10
* 25.1

juergenhoetzel:
The current implementation changes the *global* `company-minimum-prefix-length` which results in all *company-backends* kicking off.

This only happens inside strings and comments because the fsharp-completion backends only passes control to other backends when `(point)` is outside strings and comments.

I openend pull request https://github.com/fsharp/emacs-fsharp-mode/pull/104

rneatherway:
Fixed by #104

nicholas-peterson:
Bah, I didn't see the dupe issue before mine :( . Thanks for fixing this!

Issue #101, opened 2016-10-10
How to disable autocomplete drop down in comments?


michelrandahl:
_Forgive me if this is the wrong place to ask questions, I am still trying to get used to github and I don't see other places to ask questions on projects than in the issues (which I personally feel should be related to bugs, and not questions)..._

Since I updated all my emacs plugins recently, I am now getting autocompletion drop downs when writing in F# comments.
I didn't get autocompletion dropdowns in comments before updating, and I find them to be profoundly annoying since they hide my code while I'm writing on a comment. Don't get me wrong, I love having the autocompletion dropdown when working with the actual code itself, -I just don't want them when writing on my comments as they obscure the view of the code.

**Is there a way to get rid of autocompletion dropdowns when writing on comments?**

_I realize that the "autocompletion in comments" might be caused by another plugin since I am using Spacemacs which is bundled with many plugins, but as a clueless Emacs rookie I have to start the investigation somewhere..._

nosami:
Seems like you are hitting `dabbrev` completions (words that can be anywhere in the buffer).

The way the F# emacs plugin has been setup, it tries to use real intellisense first, then falls back to any other available `company-mode` (the completion engine) backends.

`(setq company-backends (list 'fsharp-ac/company-backend))` would force `company-mode` to only use real intellisense.

nosami:
The problem with ^^ is that that would clobber the company-backends for all file types. If you want to remove word completions from all file types then this might work for you (untested)

```elisp
  (setq company-backends
         (let ((backends-to-remove '(company-dabbrev company-dabbrev-code company-keywords)))
               (remove-all #'(lambda (backend) (member backend backends-to-remove)) company-backends)))

(defun remove-all (predic seq &optional res)
   "Remove items that could be anywhere inside a nested list"
   (if (null seq)
       (reverse res)
       (cond ((and (not (null (car seq))) (listp (car seq)))
              (remove-all predic (cdr seq)
                          (cons (remove-all predic (car seq)) res)))
             ((funcall predic (car seq))
              (remove-all predic (cdr seq) res))
             (t (remove-all predic (cdr seq) (cons (car seq) res))))))
 ```

michelrandahl:
Thank you so much!
Using `(setq company-backends (list 'fsharp-ac/company-backend))`  is very satisfying for me for now, since I currently only work with F# projects.

[...]

PR #104, opened 2016-10-18
Dont change `company-minimum-prefix-length'

juergenhoetzel:
Instead return (prefix . t) in prefix command to let company-mode accept
prefix lengths less than `company-minimum-prefix-length'.

This prevents premature completion of other backends (like dabbrev).

Refs #102 and #101.


rneatherway:
Fixes it for me, thanks!

PR #105, opened 2016-10-19
Don't send trailing newline to fsautocomplete

juergenhoetzel:
This fixes invalid flycheck lint messages about trailing whitespace.

empty-line

### DIFF
--- a/fsharp-mode-completion.el
+++ b/fsharp-mode-completion.el
@@ -140,7 +140,7 @@ since the last request."
         (fsharp-ac--log (format "Parsing \"%s\"\n" file))
         (process-send-string
          (fsharp-ac-completion-process (fsharp-ac--hostname file))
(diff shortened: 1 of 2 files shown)


juergenhoetzel:
I have set `(setq indicate-empty-lines t)` to indicate the EOL.

rneatherway:
I've seen this too. Thing is that `<<EOF>>` has to be on its own line right now: https://github.com/rneatherway/FsAutoComplete/blob/master/src/FsAutoComplete/CommandInput.fs#L78

So does that definitely work in all cases?

juergenhoetzel:
> I've seen this too. Thing is that ```<<EOF>>``` has to be on its own line right now:

Good catch. I also wondered why all of my *F#* files have a final newline.
Actually  *fsharp-mode* sets ```require-final-newline``` to ```t```.
In commit dd49162 I made this even stricter by ensuring the newline is present when visiting files.

PR #109, opened 2016-10-25
Define inferior-fsharp-mode as variant of comint mode

### LINKED ISSUE #108: F# Interactive buffer not displayed initial
### Description

When evaluating F# code via <kbd>C-c C-r</kbd> the first time (before an inferior F# process exists) 
an inferior F# process is created but the  corresponding process buffer is now shown. 

### Repro steps

1. Start Emacs

2. Load an .fs file

3. `fsharp-eval-region`

### Expected behavior

Inferior F# *Comint* buffer is displayed showing the result of the eval

### Actual behavior

Inferior F# *Comint* buffer is not visible until the second evaluation of F# code

juergenhoetzel:
Fixes #108

### DIFF
--- a/inf-fsharp-mode.el
+++ b/inf-fsharp-mode.el
@@ -52,20 +52,16 @@
 
 ;; Augment fsharp mode, so you can process fsharp code in the source files.
 
-(defun inferior-fsharp-mode ()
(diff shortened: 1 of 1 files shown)


rneatherway:
I couldn't actually reproduce this behaviour, but the new way works fine and is more elegant.

PR #116, opened 2016-11-15
Fix Flycheck; don't busy wait in fsharp-doc-mode

drvink:
Flycheck's idle timer will attempt to check the buffer after a change, and it
requires its callback to be called when it does this.  We were not respecting
this fully, which caused Flycheck to stall/break.  The bug was due to an
interaction between the echo area typesig/symboluse lookups and Flycheck
attempting to perform a check for which it would never get a reply.  The issue
was that Flycheck tries to run a check via post-command-hook, so if the user
merely moved the cursor but did not enter text, Flycheck would try to run a
check and never get a reply because fsharp-ac-parse-current-buffer would
decide to not send anything to FSAC (it only runs if the buffer's tick has
changed or if fsharp-ac--last-parsed-buffer is not the current buffer).  As a
result, there would be no Flycheck-related output for the FSAC process filter to
read, and Flycheck's callback wouldn't get called.


rneatherway:
Thanks!

PR #118, opened 2016-12-31
Don't change company-idle-delay

juergenhoetzel:
@nosami  I don't know about the rationale setting this to `0.03` secons in `fsharp-mode` buffers.

I often get a disruptive completion popup while  still typing. It is exacerbated by the fact that we use a `min-prefix-length` of `0`.

I would prefer not changing this `defcustom`  and let Users customize it if they want faster/slower settings.

Any options?

:fireworks:

### DIFF
--- a/fsharp-mode.el
+++ b/fsharp-mode.el
@@ -224,7 +224,6 @@
           company-backends
           company-auto-complete
           company-auto-complete-chars
-          company-idle-delay
           company-require-match
           company-tooltip-align-annotations))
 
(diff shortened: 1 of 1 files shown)


nosami:
I just wanted it to be fast out of the box. I seem to remember the default being pretty slow. Haven't seen any issues with it myself.

juergenhoetzel:
> I seem to remember the default being pretty slow. Haven't seen any issues with it myself.

I see. But the default is [0.5 seconds](https://github.com/company-mode/company-mode/blob/master/company.el#L562): if you didn't customize. No so slow for me.

Following the [Principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) i would prefer to not change this programmatically, because it will overwrite users customization.

A happy New Year to all F#/Emacs users :fireworks:

nosami:
Happy to go with the consensus :)

rneatherway:
I agree with this change.

PR #120, opened 2017-01-20
Fix FSAC hanging issue

### LINKED ISSUE #119: FSAC communication stalls indefinitely (prevents type info, Flycheck, etc.)
### Description

Flycheck checking freezes basically the moment you begin typing.

### Repro steps

Please provide the steps required to reproduce the problem

1. Type stuff.

### Expected behavior

For it to work

### Actual behavior

`FlyC*` forever!

### Known workarounds

If only. Sorry--I've spent a day on it and this time I just don't have time to debug it, and in fact, I'm not sure fsharp/emacs-fsharp-mode#116 ever actually fixed it. :(

### Related information 

* OS: happens on both Windows and Linux
* Emacs: 25.1
* Runtime: either .NET or Mono

drvink:
The protocol chatter sent to FSAC to parse some code goes like this:

```
parse "filename"
...F# code...
...F# code...
...F# code...
<<EOF>>
```

`<<EOF>>` is not an actual EOF, but literally the string `<<EOF>>`, which must be
preceded by a newline and followed by a newline.

The issue was that typing at the end of the buffer would cause the buffer's
contents to be sent to FSAC, but because the last byte of the buffer would not
be the conventional appended-at-save newline, FSAC would be sent this instead:

```
parse "filename"
...F# code...
...F# code...
...F# code...<<EOF>
```

...and thus continue waiting for input indefinitely, causing things like type
info and Flycheck to stop working.

Fixes #119.

### DIFF
--- a/fsharp-mode-completion.el
+++ b/fsharp-mode-completion.el
@@ -146,7 +146,7 @@ since the last request."
         (fsharp-ac--log (format "Parsing \"%s\"\n" file))
         (process-send-string
          (fsharp-ac-completion-process (fsharp-ac--hostname file))
-         (format "parse \"%s\" %s\n%s<<EOF>>\n"
+         (format "parse \"%s\" %s\n%s\n<<EOF>>\n"
                  (fsharp-ac--localname file)
                  (if force-sync " sync" "")
                  (buffer-substring-no-properties (point-min) (point-max)))))


rneatherway:
Good call, so I guess I never tried a file without a trailing newline. How unlucky.

PR #123, opened 2017-02-07
[WIP] Enables loading and processing project.json files currently used in .NET core

amcguier:


rneatherway:
I'm not really up to date with this stuff but I thought the project.json format had gone away again?

amcguier:
It is going away again, but all the documentation I've been able to find points to the "new" csproj format without an updated version for fsproj. Plus the dotnet cli tool for the current preview only generates project.json for F# 😢 

I'm happy to close this in favor of future compatibility, or make a second pass through and undo my changes as a PR once Microsoft settles on a format. I've been playing with AWS lambda which requires coreCLR so either way I'll maintain my own branch so I don't have to switch editors. 😄

amcguier:
Following up on this, if this is to be merged, I should maybe swap the compile command to run `dotnet build` rather than compile, although I'm not in love with that solution.

fsharp-mode.el
```diff
@@ -404,8 +404,15 @@ folders relative to DIR-OR-FILE."
                               (file-name-directory dir-or-file)))
 
 (defun fsharp-mode/find-fsproj (dir-or-file)
-  (fsharp-mode-search-upwards (rx (0+ nonl) ".fsproj" eol)
-                              (file-name-directory dir-or-file)))
+  (let ((dir-name (file-name-directory dir-or-file)))
+    (or (fsharp-mode/find-project-json dir-or-file)
```
rneatherway: I think this call and the method definition below can be removed right? It looks like you have inlined the search upwards into this method.

rneatherway:
Well, I am just not familiar with these developments so I am happy to go with your suggestion. We can always remove it again later.

amcguier:
@rneatherway I was talking to some folk in the FSharp slack and it looks like the new fsproj file is in preview 3. I'm going to test my PR against that version too, and add some tests to cover the project.json case this weekend.

rneatherway:
Sounds great, thanks.

amcguier:
looks like the fsproj updates area already out making this irrelevant. I'm closing it.

PR #125, opened 2017-04-03
Small fixes to try to prevent fsharp-mode to freeze all emacs

kyoDralliam:
Hello there,

I am observing some serious freezes when using this F# mode on a medium sized project (F*) under linux with mono and intellisense activated. Here are 2 small fixes which should improve the situation : 
- The first one remove undo-history from the background fsharp-complete buffer by default. I can't see any valid reason to maintain this and It can grow very big.
- the second is a guard to prevent fsharp mode to hang on 0-sized answer from the background process.

### DIFF
--- a/fsharp-mode-completion.el
+++ b/fsharp-mode-completion.el
@@ -332,7 +332,7 @@ If HOST is nil, check process on local system."
     (if (file-exists-p fsac)
 	(let ((proc (apply 'start-file-process
 			   fsharp-ac--completion-procname
-			   (get-buffer-create (generate-new-buffer-name "*fsharp-complete*"))
+			   (get-buffer-create (generate-new-buffer-name " *fsharp-complete*"))
 			   fsharp-ac-complete-command)))
 	  (sleep-for 0.1)
 	  (if (process-live-p proc)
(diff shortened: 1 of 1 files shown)


kyoDralliam:
I still have some errors when I start using the mode : 
```
Error: /home/kyod/fstar/FStar/src/typechecker/FStar.TypeChecker.Env.fs is not part of the loaded projects.
Loaded F# project ’typechecker.fsproj’
error in process filter: flycheck-fsharp-handle-errors: Wrong type argument: sequencep, #s(hash-table size 65 test equal rehash-size 1.5 rehash-threshold 0.8 data ("File" "/home/kyod/fstar/FStar/src/typechecker/FStar.TypeChecker.Env.fs" "Errors" nil))
error in process filter: Wrong type argument: sequencep, #s(hash-table size 65 test equal rehash-size 1.5 rehash-threshold 0.8 data ("File" "/home/kyod/fstar/FStar/src/typechecker/FStar.TypeChecker.Env.fs" "Errors" nil))
```

rneatherway:
Sorry I missed this one. Sounds like you are using this with latest FSAC? I have updated the mode today to work with that. These both seem sensible changes so I will roll them into the next release

PR #128, opened 2017-04-19
Five minute tooltip timeout seems a tad long.

darrik:
`pos-tip-show` function in pos-tip v0.4.5+ says that timeout is in seconds, and I think 300 is a bit on the long side. The default of five seconds seem more reasonable.

### DIFF
--- a/fsharp-mode-completion.el
+++ b/fsharp-mode-completion.el
@@ -752,7 +752,7 @@ has requested a popup tooltip, display a popup."
 
 (defun fsharp-ac/show-popup (str)
   (if (display-graphic-p)
-      (pos-tip-show str nil nil nil 300)
+      (pos-tip-show str)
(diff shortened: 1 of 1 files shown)


rneatherway:
I originally set it to this with the intention that the tooltip would remain present for as long as you wanted to read it. Doesn't this change mean that the documentation will disappear after 5 seconds?

darrik:
I see, yes it does. I changed it because nothing will remove the tooltip for me until the timeout has expired. Maybe that issue is local to me, will investigate.

rneatherway:
Ah, that is definitely not good. I haven't experienced that. What Emacs version and OS are you using? Does calling `pos-tip-show` have the same behaviour if you call it manually?

darrik:
Turns out I had three different pos-tip packages installed due to other packages including their own. I removed all but the one from melpa and now it works as intended. Sorry for wasting your time.

PR #131, opened 2017-06-02
Don't panic on malformed JSON (debug messages)

juergenhoetzel:
Instead just log parsing errors and try to parse next line. Fixes #130.

Issue #130, opened 2017-05-17
Malformed JSON "non-IL or abstract method with non-zero RVA"


fbie:
Completion and tool tip does not work at all and instead this error message is produced:

```
Error: F# completion process produced malformed JSON (non-IL or abstract method with non-zero RVA).
error in process filter: fsharp-ac-filter-output: Wrong type argument: hash-table-p, "Error: F# completion process produced malformed JSON (non-IL or abstract method with non-zero RVA)."
```

This happens every time when I move the cursor to any symbol in the open buffer. I am running GNU Emacs 26.0.50 (build 2, x86_64-pc-linux-gnu, GTK+ Version 3.18.9) of 2017-05-16 and Mono:

```
$ mono -V
Mono JIT compiler version 5.0.0 (Stable 5.0.0.100/9667aa6 Thu May 11 14:48:10 UTC 2017)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           __thread
	SIGSEGV:       altstack
	Notifications: epoll
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	LLVM:          supported, not enabled.
	GC:            sgen (concurrent by default)
```

rneatherway:
What version of fsharp-mode and how did you install?

fbie:
fsharp-mode version is 20170416.335, I installed via MELPA and use-package.

rneatherway:
Thanks, please can you try one or both of:

1. Troubleshooting instructions at https://github.com/fsharp/emacs-fsharp-mode#fsharp-ac-debug
2. Running the fsautocomplete.exe binary directly and seeing if that gives more information. I haven't tested with mono 5.0, so perhaps there is a bug in it.

nosami:
I've seen these messages in the console when running F# apps `non-IL or abstract method with non-zero RVA`. Not sure what's causing it yet, but I think you are going to need to filter those out from stdout

rneatherway:
Mono 5.0 is printing those to stdout?? How terrible.

nosami:
Yes. Looks like the message is coming from the compiler somehow https://github.com/fsharp/fsharp/blob/5a942daf930d7023c78f578ec6876b86a6f13f1e/src/absil/ilread.fs#L2358

oivvio:
[...]

PR #133, opened 2017-06-03
Font tidy

Gastove:
I found a couplefew more little bugs with my font locking revamp! This fixes 'em:

[||] <- the first pipe was being locked when it oughtn't be

<| <- was not being locked at all.


Gastove:
Well, so. Ran `make test` locally, which passed; forgot to run `make test-all` which very clearly doesn't. However, the failures seem to be about completion? I sure... don't think I touched completion. 

If anyone has an intuition about what the heck has gone sideways here, I'd love a Hint in the right direction. I'll get digging, see if I can figure out what's happened.

rneatherway:
Hey don't worry about those other tests, it can't have been you that broke
it. I'll take a look.

juergenhoetzel:
> Hey don't worry about those other tests, it can't have been you that broke
> it. I'll take a look.

According to travis output
```
Error: F# completion process produced malformed JSON (non-IL or abstract method with non-zero RVA
non-IL or abstract method with non-zero RVA
```

this is caused by [Malformed JSON "non-IL or abstract method with non-zero RVA"](https://github.com/fsharp/emacs-fsharp-mode/issues/130), fixed by [Don't panic on malformed JSON (debug messages) ](https://github.com/fsharp/emacs-fsharp-mode/pull/131)

Gastove:
AH. Okay, cool -- I didn't have time to look at the other bug reports/errors. Thanks, @juergenhoetzel

rneatherway:
Thanks for the PR, 1.9.7 including it should be live on MELPA by tomorrow.

PR #136, opened 2017-06-10
Use correct F# interactive prompt regex

juergenhoetzel:
I suppose this is a leftover from the original OCaml code.

I still work on [Org Babel](http://orgmode.org/worg/org-contrib/babel/) support. This is fix  required to let ```org-babel-comint-with-output``` dissect the output and also for  `comint` navigation.


juergenhoetzel:
Also did some code cleanup in 1783bb7 and c9fe0c6 :beer:

Issue #138, opened 2017-06-20
Is there a way to disable all functionality (including flycheck) that relies on background processes?


preetpalS:
### Description

Basically I just want syntax highlighting and indentation without any other interactivity that relies on background processes.

### Expected behavior

Ideally you would set a variable and disable all interactivity (no flycheck, no automatic background process related errors on hovering over anything).

### Actual behavior

Even if you have `fsharp-ac-intellisense-enabled` set to nil, you will constantly get the error "Error: background intellisense process not running".

### Known workarounds

None that I know of.

rneatherway:
There currently isn't such a way. The variable you mention is a sensible place to do it, by adding an `if` here: https://github.com/fsharp/emacs-fsharp-mode/blob/master/fsharp-mode.el#L272

rneatherway:
Fixed by #139

PR #139, opened 2017-06-21
Disable flycheck and fsharp-doc-mode when fsharp-ac-intellisense-enab…

preetpalS:
…led is set to nil.

This is a fix for issue #138. Note that I am disabling **fsharp-doc-mode** as well when `fsharp-ac-intellisense-enabled` is set to `nil` since I believe it relies on a background intellisense process as well.

I ran the test suite before and after the change (`gmake test-all`) on FreeBSD 11 and all tests passed in both runs. I also used the mode once and the constant "Error: background intellisense process not running" messages were not appearing.

### DIFF
--- a/fsharp-mode.el
+++ b/fsharp-mode.el
@@ -269,8 +269,10 @@
       (setq compile-command (fsharp-mode-choose-compile-command file))
       (fsharp-mode--load-with-binding file)))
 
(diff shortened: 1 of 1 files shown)


rneatherway:
Thanks perfect!

PR #142, opened 2017-09-06
Add support for using build.sh scripts

chamook:
This adds a function to find a build.sh file in the same way a sln or
fsproj file is found, and then uses that to build if it's available.

I've never really done anything with emacs lisp before, so I don't know if this is really dumb or not, but it seems to work for me.


juergenhoetzel:
> This adds a function to find a build.sh file in the same way a sln or
> fsproj file is found, and then uses that to build if it's available

Thanks. 
But wouldn't we need additional checks for Windows (`build.sh`, `build.ps1`)?
I wonder if the `cond` clause will ever match: I use [ProjectScaffold](https://github.com/fsprojects/ProjectScaffold) and `dotnet new` based Projects.


@rneatherway btw: I wonder why we use `fsharp-mode-search-upwards` for locating files instead of using Emacs builtin `locate-dominating-file`

chamook:
I tested this on Windows using projects created with [Forge](http://forge.run/) which IME creates both ```build.sh``` and ```build.cmd```. I'm happy to expand it to search through several different extensions if you think it's a worthwhile addition :)

rneatherway:
I think that `build.cmd` also would be useful.

I think we invented our own function because we weren't aware of the builtin.

rneatherway:
Thanks for the update, it looks good, but unfortunately breaks some of the unit tests as I guess this is a behaviour change. Could you take a look at those please? Are you happy with the changes?

chamook:
Sorry for the slow response.

I'm ok with the code as it stands, I was hoping to find a solution that was more similar to the "fsproj" finder but that didn't work out.

I've added a test to cover finding a build script, but I'm not sure how to go about solving the other test failures. It seems to be caused by Travis having a build.sh file in a parent directory to the tests. When running tests locally I don't get any failures, unless I introduce a build.sh file somewhere up the directory tree. I could remove the existing tests, but that doesn't sit right with me.  I'm going to look into this further to try to find a better solution, but I wanted to provide an update so that you don't think I've abandoned this, and also to ask if you know a smart way of handling this?

rneatherway:
Yes, I can reproduce that here by running `touch ../build.sh` before running the tests. I suppose it is because build.sh is earlier in the precedence order. I'm not sure how to get around this, that file will always be present on Travis. Maybe it shows that the feature (prior to your PR) is a bit flawed, and shouldn't search upwards indefinitely.

chamook:
I defined a variable for the list of scripts to search for, and then used dummy values in test to get around the presence of that script on Travis. Does that seem like an okay solution?

rneatherway:
That works for me. Any objections @juergenhoetzel ?

chamook:
I amended the commit to fix a typo I made in the doc string for the list. :$

juergenhoetzel:
> I defined a variable for the list of scripts to search for, and then used dummy values in test to get around the presence of that script on Travis. Does that seem like an okay solution?

This will still set `compile-command` to an invalid value on a Windows-System if users don't customize this list:

```
ELISP> (fsharp-mode/find-build-script default-directory)
"c:/cygwin/home/juergen/emacs/site-elisp/fsharp-mode/test/CompileCommandData/build-script/build.sh"
```

How about setting the build-script name conditionally like this:

```elisp
(defvar fsharp-mode/build-script-name
  (cond
   ((eq system-type 'windows-nt) "build.cmd")
   (t "build.sh"))
  "Build script name to use when compiling.")
```

chamook:
I run it on Windows and prefer it to call the shell script tbh - but given that it would be customizeable, I think that's a fair default to have.

juergenhoetzel:
> I run it on Windows and prefer it to call the shell script tbh - but given that it would be customizeable, I think that's a fair default to have.

I wonder how you can execute Shell scripts on Windows (without using _Linux Subsystem_  or _Cygwin_)?

chamook:
Git bash runs them fine :)

juergenhoetzel:
> Git bash runs them fine :)

I see. But we can't assume  _Git bash_ is installed on every system. 
Using a `defcustom`  Users could still use `build.sh` as `compile-command` when they have _Git bash_ installed.

> I run it on Windows and prefer it to call the shell script tbh


Are there any advantages using the `build.sh` over using `build.cmd` or `build.ps1`?
In my  experience they are just wrapper scripts around _Paket_ and _Fake_.

rneatherway:
Does that look OK @juergenhoetzel ?

juergenhoetzel:
The `build.cmd`  has to be invoked from the Project directories. When editing a file in a subdir, the `compile` command doesn't work:

```
-*- mode: compilation; default-directory: "~/fsharp/XPlot/src/XPlot.GoogleCharts/" -*-
Compilation started at Wed Sep 20 07:30:31

c:/cygwin/home/juergen/fsharp/XPlot/build.cmd
Das System kann den angegebenen Pfad nicht finden.

Compilation exited abnormally with code 3 at Wed Sep 20 07:30:31

```

chamook:
I had forgotten about that - I've been modifying my build scripts to include

    # if paket isn't here we're probably running from a project
    # subfolder, so jump up to the sln folder before running the script
    paketPath=.paket/paket.exe
    if [ ! -e $paketPath ]; then
        cd ..
    fi

but that's not really a great thing to require (and this implementation isn't very flexible).

AFAICT the issue arises because `default-directory` is set to the directory containing the current file, and that's the context in which the build script is run. My initial thought was to change directory before running the script, and then change back afterwards - but it looks like all that happens in `fsharp-mode` is setting a `compile-command` which is handled elsewhere, so I'm not sure that's practical.

I can give that a try, but do you guys know a better approach?

juergenhoetzel:
Escaping the `buildsh` and the directory is also a good idea because might contain special chars (like whitespace):

```
     (cond
      (makefile          compile-command)
      (buildsh           (format "cd %s && %s"
                               (shell-quote-argument (file-name-directory buildsh))
                               (shell-quote-argument buildsh)))
      (proj              (combine-and-quote-strings (list fsharp-build-command "/nologo" proj)))
```

chamook:
Thank you for the help :)

I tried using the ```shell-quote-argument``` function on the paths, but had issues with it escaping the ```:``` character in windows file paths which made the command invalid. I've added quotes around the paths instead, which should cover whitespace at least.

juergenhoetzel:
> Thank you for the help :)
> 
> I tried using the shell-quote-argument function on the paths, but had issues with it escaping the : character in windows file paths which made the command invalid. I've added quotes around the paths instead, which should cover whitespace at least.

Hi @chamook 

I couln't reproduce this issue. I don't see how the the colon character gets escaped:

     ELISP> (shell-quote-argument default-directory)
     "\"c:/cygwin64/home/juergen/fsharp/FsAutoComplete/src/FsAutoComplete/\""

What do you get?

chamook:
Hi, apologies for the long turnaround. When building a compile command on my windows machine using `shell-quote-argument` I get the following:

```cd c\:/development/Forge/ && c\:/development/Forge/build.cmd```

which gives me `No such directory found via CDPATH environment variable`

**Comment by juergenhoetzel** [comment]
I have setup the same environment ([Forge CLI-Tool](https://github.com/fsharp-editing/Forge)) and can't reproduce.

Could you show the output of `(describe-variable compile-command)` in a `F#` buffer? Mine:

```
compile-command is a variable defined in ‘compile.el’.
Its value is
"cd \"c:/development/Forge/\" && \"c:/development/Forge/build.cmd\""
```

Output of `(compile)`:

```
-*- mode: compilation; default-directory: "c:/development/Forge/" -*-
Compilation started at Sat Feb 10 17:02:13

cd "c:/development/Forge/" && "c:/development/Forge/build.cmd"
Checking Paket version (downloading latest stable)...
Version 5.135.0 not found in cache.
Starting download from https://github.com/fsprojects/Paket/releases/download/5.135.0/paket.exe
Done.
....
```

**Comment by chamook** [comment]
```
compile-command is a variable defined in `compile.el'.
Its value is
"cd c\\:/development/Forge/ && c\\:/development/Forge/build.cmd"
Original value was "make -k "
Local in buffer Commands.fs; global value is "make -k "
```

**Comment by juergenhoetzel** [comment]
We're coming closer to it: The directory and command parts doesn't get escaped in your `compile-command`:

`"cd c\\:/development/Forge/ && c\\:/development/Forge/build.cmd"`

Mine:

`"cd \"c:/development/Forge/\" && \"c:/development/Forge/build.cmd\""`

Could you make sure you are really using  this code in your setup:
```
(buildsh           (format "cd %s && %s"
                               (shell-quote-argument (file-name-directory buildsh))
                               (shell-quote-argument buildsh)))
```

Maybe you are using outdated byte-compiled `fsharp-mode.el` ?

juergenhoetzel:
*ping* :rabbit:

juergenhoetzel:
ping

Issue #140, opened 2017-08-24
.NET Core 2 support?


julienXX:
### Description

Keep getting 'file not part of the project' on a brand new .NET Core 2 project.
For a classlib:
```
The file ’/Users/julien/Code/Examples/Demo/Demo.Core/Library.fs’ was not part of the project. Did you call InvalidateConfiguration when the list of files in the project changed?
```
For a XUnit test file:
```
Error: /Users/julien/Code/Examples/Demo/Demo.Tests/Tests.fs
```
Also saw this in the *Messages* buffer:
```
error in process filter: Wrong type argument: flycheck-error-p, [cl-struct-flycheck-error #<buffer  *Minibuf-1*> fsharp-fsautocomplete "/Users/julien/Code/Examples/Demo/Demo.Tests/Tests.fs" 4 6 "The namespace or module 'Xunit' is not defined." error nil]
```

### Repro steps

Create a .NET Core 2 classlib.

### Expected behavior

It should work as in .NET Core 1 projects.

### Actual behavior

See Description.

### Known workarounds

Changing the type of the classlib project from `netstandard2.0` to `netcoreapp2.0` works.

### Related information 

* OSX Sierra
* Emacs 25.1
* .NET Core 2.0

With debug logs enabled I get:
```
11:40:12.732619: project "/Users/julien/Code/Examples/Demo/Demo.Tests/Demo.Tests.fsproj" verbose
[...]

PR #143, opened 2017-09-14
Update to FSAC 0.34.0

rneatherway:
Still needs manual testing, automated tests passed locally which I found a bit surprising.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ load_integration_tests = $(patsubst %,-l %, $(integration_tests))
 # Autocomplete binary distribution.
 ac_name    = fsautocomplete
 ac_exe     = $(bin_d)/$(ac_name).exe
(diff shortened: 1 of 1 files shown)


enricosada:
ref #140 

@rneatherway api wise is the same.
afaik changes ares:
- command line arg parser (not commands) is changed (now use Argu), but i tried to maintain exact same previous args. `--help` show current. (some args require `=`, like before)
- options now are serialized as `null | x` instead of `null | "Some": x ` , using same convention as http mode
  - lint response is affected by new option
  - new project info are affected, but emacs doesnt use these, so np. previous are ok
- new specific error info, but anyway are `error`, so is the same, the continue to have message

rneatherway:
Thanks! The linter is working locally so I will merge, I guess that field wasn't used.

PR #153, opened 2017-10-28
Add access control keywords to declaration regexes

Tarmil:
This enhances both coloring (`let private f x` now colors `f` as a function) and imenu support (`f` is listed, not `private`).


rneatherway:
Thanks!

PR #163, opened 2018-03-28
Use emacs-travis.mk instead of deprecated evm

juergenhoetzel:
Also add emacs-26-pretest and emacs-snapshot to build matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,40 @@
-language: csharp
+language: emacs-lisp
+# Emacs fails to build in container-based builds on Travis
+# See https://github.com/travis-ci/travis-ci/issues/9061
+# and https://github.com/moby/moby/issues/22801
 sudo: required
-
-before_install:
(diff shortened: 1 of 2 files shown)


juergenhoetzel:
Need fix (dont use internal `tramp-make-tramp-file-name`) for Emacs-26 compatibility: http://lists.gnu.org/archive/html/tramp-devel/2017-09/msg00001.html

```
In toplevel form:
fsharp-mode-completion.el:211:10:Error: tramp-make-tramp-file-name called with 4 arguments, but requires 6-7
make: *** [fsharp-mode-completion.elc] Error 1
```

juergenhoetzel:
> Need fix (dont use internal tramp-make-tramp-file-name) for Emacs-26 compatibility: http://lists.gnu.org/archive/html/tramp-devel/2017-09/msg00001.html

Fixed in fad843b

PR #181, opened 2018-07-11
Enable integration tests on Windows

### LINKED ISSUE #12: Emacs tests don't work on Windows
_From @rneatherway on October 1, 2014 11:46_

A few different bugs/reasons, mostly upstream.
1. Byte compilation check. This fails when using Chocolatey Emacs (including AppVeyor) because the shim executable in the Chocolatey bin directory affects string escaping. Example:
   
   ```
   emacs --batch --eval "(message \"hello\")
   ```
   
   This works normally, but with Chocolatey emacs we get `undefined symbol hello`. This affects the setting of the load path with `--eval`. Currently can work around by launching the real Emacs binary directly, or using rather bizarre string escaping like `emacs --batch --eval "(message \"\\\"hello\\\"\")`. Could also try to move the load-path modification into a function so we don't need to mention strings on the command line.
2. Unit tests fail because Emacs on windows doesn't seem able to download ELPA packages in batch mode:
   
   ```
   >c:\utils\emacs\bin\emacs -Q --batch -f toggle-debug-on-error --eval "(setq package-archives '((\"gnu\" . \"http://elpa.gnu.org/packages/\")))" --eval "(package-refresh-contents)"
   Debug on Error enabled globally
   Contacting host: elpa.gnu.org:80
   Memory exhausted--use C-x s then exit and restart Emacs
   ```
3. Integration tests fail because the background process doesn't seem to get killed cleanly after the first test. Description copied from #728:
   
   ```
   Running 2 tests (2014-09-29 23:35:38+0100)
   Loading vc-git...
   passed  1/2  check-completion
   Test check-project-files backtrace:
   process-send-string(#<process fsharp-complete> "quit\n")
   log-psendstr(#<process fsharp-complete> "quit\n")
   
   ... <snip> ...
   
   Test check-project-files condition:
   (file-error "writing to process" "invalid argument" #<process fsharp-complete>)
   FAILED  2/2  check-project-files
   ```
   
   I have observed that fsautocomplete still seems to be running at the beginning of the second test, while it should have been stopped by the cleanup of the first.
   
   It seems like it may be related to an elpy bug (https://github.com/jorgenschaefer/elpy/issues/212), which in turn seemed to be an Emacs bug (http://debbugs.gnu.org/cgi/bugreport.cgi?bug=18396). However, this would only happen if sufficiently large quantities of data were being sent in both directions at once. As emacs doesn't seem to hang, it's less likely.

_Copied from original issue: fsharp/fsharpbinding#748_

juergenhoetzel:
Fixes #12.

### DIFF
--- a/.appveyor/test.bat
+++ b/.appveyor/test.bat
@@ -1,5 +1,0 @@
-set HOME=%~dp0\tmp
-
-emacs -l test\test-common.el --batch --eval "(progn (set-default-coding-systems 'utf-8) (load-packages))"
-
-emacs -l test\test-common.el --batch --eval "(progn (set-default-coding-systems 'utf-8) (run-fsharp-unit-tests))"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,6 +15,6 @@ install:
   - ps: scoop bucket add extras
   - ps: scoop install emacs
 build_script:
-  - cmd: .\.appveyor\test.bat
+  - cmd: make test-all
 test: off
  


juergenhoetzel:
Duplicate:the topic is addressed in #226 now

PR #189, opened 2019-01-24
Fix a small bug in font locking for active patterns

Gastove:
This commit fixes a bug in which only the final name in a multi-symbol active
record would be font locked. The trouble is that the regexp requires the closing
banana clip |) as part of the pattern; the final paren should be _optional_.
Thus, the only meaningful change is adding a ? inside the regexp. A newline is
introduced to make the whole thing slightly more readable.

### DIFF
--- a/fsharp-mode-font.el
+++ b/fsharp-mode-font.el
@@ -293,7 +293,8 @@ with initial value INITVALUE and optional DOCSTRING."
       (,fsharp-pattern-function-regexp 1 font-lock-function-name-face)
       ;; Active records
       ("(|" (0 'fsharp-ui-operator-face)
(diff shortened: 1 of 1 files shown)


juergenhoetzel:
> The build failure appears unrelated to my work. Is this a known thing that should be sorted in its own branch, or a "just-me" I should dig in to here?

I also experienced the travis timeout issue. I will check out the [travis_retry](https://docs.travis-ci.com/user/common-build-problems/#timeouts-installing-dependencies) feature to mitigate the effect of network timeouts.

Gastove:
Thanks, Jürgen :+1:

juergenhoetzel:
Thanks!

> Active _patterns_, not active records. Bleh ;P

No worries. The typo was also in the Source comments :laughing:

PR #190, opened 2019-01-27
Fix attribute locking, improve imenu

Gastove:
This PR does two things:

1. Updates `imenu` generation so that elements in the `imenu` display aren't all just labelled "Function", while adding support for jumping to Module declarations.
2. Fixes font locking for parameterized attributes. 

Before, locking an attribute like Argu's `[<AltCommandLine("-d")>]` would fail, as the  current locking cannot recognize either the internal parens or the internal string. This PR fixes that, locking the string as such, and making sure everything else is correctly, er, attributed. 

I do feel... a little sheepish about putting both things on the same branch. Please let me know if two PRs are preferred, or if these changes are minor/cosmetic enough that it's no bother.

### DIFF
--- a/fsharp-mode-font.el
+++ b/fsharp-mode-font.el
@@ -288,7 +288,7 @@ with initial value INITVALUE and optional DOCSTRING."
       ;; attributes
       (,fsharp-attributes-regexp
        (1 font-lock-preprocessor-face)
-       (2 font-lock-string-face)
+       (2 font-lock-string-face nil t)
        (3 font-lock-preprocessor-face))
       ;; ;; type defines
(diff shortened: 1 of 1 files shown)


juergenhoetzel:
> I do feel... a little sheepish about putting both things on the same branch. Please let me know if two PRs are preferred, or if these changes are minor/cosmetic enough that it's no bother.

Yes, this is preferred because it's easier to reason about and test your fixes/improvement.
BTW: Any tests are also welcome :smile:   

I merged/rebased to `master` 

Thanks a lot for your regex-magic :+1:  and help for `fsharp-mode`

Issue #183, opened 2018-08-01
Can't load .Net Core projects that reference other projects


bretthall:
### Description

If I have a .Net Core project that references another project then I get an error when trying to load the project.

### Repro steps

1. Clone https://github.com/bretthall/FsEmacsProblem

2. Open Prog/Program.fs in emacs 

### Expected behavior

Autocomplete and everything else should work.

### Actual behavior

`parsing resolved p2p refs, expected properties not found ’[|"ProjectReferenceFullPath=c:\projects\FsEmacsTest\Lib\Lib.fsproj;"|]’` will appear in `*messages*` and anything that relies on the project file being loaded (e.g. auto-complete) will not work. Note that Prog.fsproj has a reference to Lib.fsproj. Opening a file from the Lib project (which doesn't reference another project) works as expected. Both projects were created and the reference created using the dotnet CLI tool.

### Known workarounds

None known.

### Related information 

* Operating system: Windows 10 and OsX
* Branch: Version 20180518.1120 from melpa
* Emacs version: 25.3.1 and 26.1
* .NET Runtime, CoreCLR or Mono Version: .Net Core 2.1

juergenhoetzel:
> parsing resolved p2p refs, expected properties not found ’[|"ProjectReferenceFullPath=c:\projects\FsEmacsTest\Lib\Lib.fsproj;"|]’ will appear in *messages* and

I hit the same issue. This is fixed in `fsautocomplete` master. We have to wait for a new release. Upstream Bug: [New release 0.35?](https://github.com/fsharp/FsAutoComplete/issues/291) :cry:

bretthall:
Bummer, guess I'll just keep an eye out for the next release then. I tried building fsautocomplete myself off the branch the Ionide is using, but I keep getting crashes when emacs tries to use it. Maybe I'll have some more time to look at it later.

juergenhoetzel:
> Bummer, guess I'll just keep an eye out for the next release then. I tried building fsautocomplete myself off the branch the Ionide is using, but I keep getting crashes when emacs tries to use it. Maybe I'll have some more time to look at it later.

`master` branch worked for me. I hope `fsautocomplete` gets a new maintainer soon.

ed-ilyin:
So, is it possible to do the new release of fsharp-mode using current master of fsautocomplete?

wanderley:
Does someone find a workaround for this issue?

paytonrules:
[...]

PR #192, opened 2019-02-25
Add the option to build and install with different versions of fsac

jdodds:
This should be helpful for people running into #183 and related issues, and seems like a useful thing to be able to do in general.


juergenhoetzel:
> This should be helpful for people running into #183 and related issues, and seems like a useful thing to be able to do in general.

Indeed: Very useful! Thanks a lot for your contribution!

PR #201, opened 2019-04-19
[WIP] Support .NET Core as the FSAC runtime.

TOTBWF:
This is very WIP, but it aims to support .NET Core as a runtime for FSAC. This would remove `mono` as a dependency for those that choose to.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.elc
 bin
+bin_netcore
 tmp
 
 # Useful for doing releases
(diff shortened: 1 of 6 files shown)


TOTBWF:
Right now, certain projects work, but others (mysteriously) fail with either
```

TOTBWF:
Right now, the biggest issue I'm facing is that when there multiple `TargetFrameworks`, `FsAutoComplete` will try to build using the wrong `TargetFramework` and fail. It looks like fsharp/FsAutoComplete#347 ought to fix this.

ebellani:
hey folks, is there anything missing here that I can help with? I am interested in being able to use emacs as my dev tool for f# in .net core.

TOTBWF:
I've put this on the back burner in favor of supportinghttps://github.com/emacs-lsp/lsp-mode, which I would highly recommend. Feel free to ask me any questions about setting it up 😀

Issue #204, opened 2019-05-17
Fails loading project created with "dotnet new classlib -lang f#"


kqr:
### Description

I cannot get fsharp-mode to load any project file: the simplest I've tried so far has been the one generated by `dotnet new classlib -lang f#` which also fails. 

### Repro steps

(1) Create a directory with the following two files:

`Library.fs`:

```
namespace fsharp_mode_repro

module Say =
    let hello name =
        printfn "Hello %s" name
```

`fsharp_mode_repro.fsproj`:

```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>netstandard2.0</TargetFramework>
    <RootNamespace>fsharp_mode_repro</RootNamespace>
  </PropertyGroup>

  <ItemGroup>
    <Compile Include="Library.fs" />
  </ItemGroup>

</Project>
```

(2) Open `Library.fs` in Emacs.

(3) For good measure, press `C-c C-p` and navigate to the `fsharp_mode_repro.fsproj` file, and press `RET`.

### Expected behavior

I expect the project file to be loaded.

### Actual behavior

My message buffer gets a stack trace:

```
Unexpected internal error. Please report at https://github.com/fsharp/FsAutoComplete/issues, attaching the exception information:
System.InvalidOperationException: object_op ---> System.EntryPointNotFoundException: SystemNative_RealPath
  at (wrapper managed-to-native) Interop+Sys.RealPath(string)
  at System.IO.CoreFX.FileSystemWatcher+RunningInstance.Start () [0x00000] in <e9297c7c5cde4e1285346aced9ff2658>:0 
  at System.IO.CoreFX.FileSystemWatcher.StartRaisingEvents () [0x00067] in <e9297c7c5cde4e1285346aced9ff2658>:0 
  at System.IO.CoreFX.FileSystemWatcher.StartRaisingEventsIfNotDisposed () [0x00019] in <e9297c7c5cde4e1285346aced9ff2658>:0 
  at System.IO.CoreFX.FileSystemWatcher.set_EnableRaisingEvents (System.Boolean value) [0x0001d] in <e9297c7c5cde4e1285346aced9ff2658>:0 
  at (wrapper remoting-invoke-with-check) System.IO.CoreFX.FileSystemWatcher.set_EnableRaisingEvents(bool)
[...]

PR #205, opened 2019-05-17
Add workaround for file watcher issue in FSAC on OS X

kqr:
This is the best workaround I've found so far. Without it, I cannot even load
the simplest of project files.

Fixes issue #204.

### DIFF
--- a/fsharp-mode-completion.el
+++ b/fsharp-mode-completion.el
@@ -337,20 +337,25 @@ If HOST is nil, check process on local system."
 			      (fsharp-ac--clear-symbol-uses))))
     (fsharp-ac--reset)))
 
+(defun fsharp-ac--get-process-environment ()
+  (let ((proc-env process-environment))
+    (when fsharp-ac-using-mono
+      ;; workaround for Mono = 4.2.1 thread pool bug
+      ;; https://bugzilla.xamarin.com/show_bug.cgi?id=37288
+      (let ((threads-per-cpu (getenv "MONO_THREADS_PER_CPU")))
+	(when (or (null threads-per-cpu)
+		  (< (string-to-number threads-per-cpu) 8))
+	  (push "MONO_THREADS_PER_CPU=8" proc-env)))
+      ;; Workaround for file system watcher issue in FSAC
+      (when (eq system-type 'darwin)
+	(push "MONO_MANAGED_WATCHER=disabled" proc-env)))
+    proc-env))
+
 (defun fsharp-ac--configure-proc ()
   (let* ((fsac (if (tramp-tramp-file-p default-directory)
 		   (concat (file-remote-p default-directory) (car (last fsharp-ac-complete-command)))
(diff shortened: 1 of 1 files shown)


juergenhoetzel:
> This is the best workaround I've found so far. Without it, I cannot even load
> the simplest of project files.
> 
> Fixes issue #204.

I wonder what negative effects this change could have:
```
MONO_MANAGED_WATCHER=disabled"
``` 
IIRC the watcher is used by `fsautocomplete.exe` to track changes on project files. 

It seems like a [leak of file descripors](https://github.com/fsharp/FsAutoComplete/#filewatcher-exceptions). On my Arch Linux system only 9 `inotify` file descriptors are allocated (using your sample project):
 
```
juergen@lemmy:~/fsharp/test → lsof -n -p $(pgrep -f fsautocomplete.exe)|grep inotify
mono    869 juergen    4r  a_inode   0,13        0   11770 inotify
mono    869 juergen    5r  a_inode   0,13        0   11770 inotify
mono    869 juergen    6r  a_inode   0,13        0   11770 inotify
mono    869 juergen    7r  a_inode   0,13        0   11770 inotify
mono    869 juergen    8r  a_inode   0,13        0   11770 inotify
mono    869 juergen    9r  a_inode   0,13        0   11770 inotify
mono    869 juergen   10r  a_inode   0,13        0   11770 inotify
mono    869 juergen   11r  a_inode   0,13        0   11770 inotify
mono    869 juergen   12r  a_inode   0,13        0   11770 inotify
```

Can you check this out on `macos` using similar tools? Thanks!

kqr:
Are you asking me to check for allocated file descriptors when I'm running with `MONO_MANAGED_WATCHER=disabled` or with the "managed watcher" (whatever it is)? To be clear: I cannot even start running `fsautocomplete.exe` with the managed watcher because it quits right away.

If I'm allowed to guess wildly, I would suspect it's using some sort of filesystem watcher API that either a) does not exist in modern OS X releases, or at least b) is protected by some sort of security measure. But I don't know enough about OS X to make a more educated guess than that – I'm forced to use the operating system at work, but I'm not interested enough in it to learn a lot about it...

kqr:
There is also this that I found: https://github.com/fsharp/FsAutoComplete/issues/331#issuecomment-484730743

kqr:
Turns out if I install mono from www.mono-project.com instead of homebrew (one of the more popular package managers for OS X), FSAC loads projects just fine without workaround, and running the `lsof` command you suggested shows me a bunch of `KQUEUE` entries which I think is what corresponds to inofity in the BSD/OS X world.

So I retract this pull request, but if it's possible to detect this situation (i.e. Mono installed from homebrew and exception relating to FileWatcher when trying to load project file) and throw a more helpful error message suggesting to download and install from www.mono-project.com instead, that'd be a good idea. At least if this issue with Mono from homebrew is going to be long-lived.

PR #206, opened 2019-05-17
Set default build command to msbuild if it is found

kqr:
Newer versions of Mono have deprecated xbuild and encourage usage of the msbuild
command that ship with Mono instead. Picking xbuild as the default causes issues
with modern VS project files, which it does not understand. It should be
reasonably safe to prefer using msbuild over xbuild on any platform where
msbuild is found.


juergenhoetzel:
Thanks :+1:

PR #208, opened 2019-07-11
Add setting for conservative indent on first line after bracket

kqr:
If I press return when point is at the square in the following line of code:

```
let SigninButton = FunctionComponent.Of (fun (props : UntypedProps) ->▊
)
```

fsharp-mode helpfully indents for alignment:

```
let SigninButton = FunctionComponent.Of (fun (props : UntypedProps) ->
·········································▊
)
```

I prefer a less aligning indentation style, where I indent only by the same
amount any time it is legally possible to do so. In this case, what I wanted was
the more conservative

```
let SigninButton = FunctionComponent.Of (fun (props : UntypedProps) ->
····▊
)
```

This commit adds a customizeable variable that makes fsharp-mode indent in this
more conservative way instead.

### DIFF
--- a/fsharp-mode-indent.el
+++ b/fsharp-mode-indent.el
@@ -53,6 +53,14 @@ statement are given this extra offset."
   :type 'integer
   :group 'fsharp)
 
+(defcustom fsharp-conservative-indentation-after-bracket nil
+  "Indent by fsharp-continuation-offset also after an opening bracket.
+The default indentation depth on a new line after an opening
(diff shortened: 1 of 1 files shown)


juergenhoetzel:
Thanks :+1: 
I hope that after [switching to the stdio protocol](https://github.com/fsharp/FsAutoComplete/issues/410) I can find the time to  add support for an external F# code formatter.

PR #209, opened 2019-07-11
Add curly brace as token that should be outdented

### LINKED ISSUE #203: AutoIndentation is broken for computation expressions
When writing computation expressions, auto-indentation produces some surprising results.

For example, typing
```
let foo =
    async {
           return 0
        }
```
results in
```
let foo =
    async {
        return 0
        }
```

Also, typing
```
let foo = async {
        return 0
    }
```
results in
```
let foo = async {
    return 0
    }
```

kqr:
There are three examples I know of what this does:

Before this patch, we see the following three indentations happening:

```
type X =
····{
········field : WithType
········}
```

```
let foo =
····async {
········field : WithType
········}
```

```
let foo = seq {
········field : WithType
········}
```

Whereas with this patch, we instead get

```
type X =
····{
········field : WithType
····}
```

```
let foo =
····async {
········field : WithType
····}
```

```
let foo = seq {
····field : WithType
}
```

This, I believe, fixes #203.

### DIFF
--- a/fsharp-mode-indent.el
+++ b/fsharp-mode-indent.el
@@ -156,7 +156,8 @@ as indentation hints, unless the comment character is in column zero."
                              "finally"
                              "end"
                              "done"
-                             "elif")
+                             "elif"
+                             "}")
                            "\\|")
           "\\)")
   "Regular expression matching statements to be dedented one level.")


kqr:
I have since noticed that this may not fix some other undesirable indentation levels when the nesting is deeper. I can't promise to find time to address that in full.

juergenhoetzel:
Thanks! :+1:

PR #212, opened 2019-07-25
Corrected links to Forge and MELPA

MiyamotoAkira:
Forge was a 404.
Melpa had a redirect, but now is pointing to the canonical one with https.


juergenhoetzel:
Thanks! :+1:

PR #226, opened 2019-11-26
WIP: Adding new requirements for appveyor installation

MiyamotoAkira:
Start the process of getting the new cask/buttercup system running on appveyor

### DIFF
--- a/.appveyor/test.bat
+++ b/.appveyor/test.bat
@@ -1,5 +1,6 @@
 set HOME=%~dp0\tmp
 
-emacs -l test\test-common.el --batch --eval "(progn (set-default-coding-systems 'utf-8) (load-packages))"
+REM emacs -l test\test-common.el --batch --eval "(progn (set-default-coding-systems 'utf-8) (load-packages))"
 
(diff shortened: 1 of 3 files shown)


MiyamotoAkira:
Tests running now. Four of them fail (also in local), and still have to fix that the end of the run should be based on the output of the commands

juergenhoetzel:
> Tests running now. Four of them fail (also in local), and still have to fix that the end of the run should be based on the output of the commands

Thanks for working on this :+1: 
I wonder why Appveyor reports succees. Do we need something like `bash` `set -o pipefail` here:
```bash
cask exec buttercup -L . -L (Join-Path (Split-Path $PSScriptRoot -parent) "test") | Out-String
```

MiyamotoAkira:
Could be because of Out-String. Between today and tomorrow i should be able
to get that sorted (he says hopefully), and then investigate the four
failures

juergenhoetzel:
Appveyor installation is  obsoleted by the #249 

The windows ci-job is a bad hack because of the cask-windows problems: Any support to improve it [eldev](https://github.com/doublep/eldev) is welcome :smile:

Issue #231, opened 2019-12-07
Script files(.fsx) not working


FsharperOnTheRun:
### Description

No completion, type signature on F# script files(.fsx). 
Code files(.fs) works as expected.

### Repro steps

1. Step A

open a simple script file, Foo.fsx:
```fsharp
let x = 10
x.ToString()
x.T
```

2. Step B
- M-x eglot
- highlight x, ToString
- try complete x.T

### Expected behavior

- x, ToString type signatures should show on mini-buffer
- x.T should display completions

### Actual behavior

"flymake:Wait" on mode line never goes away.

And Error:
```lisp
[...]
client-request (id:28) Sat Dec  7 16:17:15 2019:
(:jsonrpc "2.0" :id 28 :method "textDocument/completion" :params
      (:textDocument
       (:uri "file:///home/me/Test/fsharp/core-3.0.101/hello-world/Foo.fsx")
       :position
       (:line 5 :character 3)
       :context
       (:triggerKind 1)))

server-reply (id:28) ERROR Sat Dec  7 16:17:15 2019:
(:jsonrpc "2.0" :id 28 :error
      (:code -32603 :message "no type check results"))

internal (id:24) Sat Dec  7 16:17:25 2019:
(:timed-out :textDocument/signatureHelp :id 24 :params
        (:textDocument
         (:uri "file:///home/me/Test/fsharp/core-3.0.101/hello-world/Foo.fsx")
         :position
         (:line 5 :character 3)))
```

### Known workarounds

Stick with .fs and never touch .fsx? :(

### Related information 

* Operating system
Ubuntu 19.10
* Branch
melpa version fsharp-mode-20191130.1857
* Emacs version
GNU Emacs 26.3
* .NET Runtime, CoreCLR or Mono Version
.net core 3.0.101

juergenhoetzel:
Thanks for the detailed report. I can confirm. It's most likely related to this ionide issue: [FSAC Runtime = netcore does not work with fsx
#948](https://github.com/ionide/ionide-vscode-fsharp/issues/948)

juergenhoetzel:
[4.2.0 - No auto complete or typechecking in FSX files](https://github.com/ionide/ionide-vscode-fsharp/issues/1244)

`*EGLOT (hello/fsharp-mode) stderr*` shows:
```
[...]

PR #234, opened 2019-12-13
Add .fsx completion issue workaround to README.org. Refs #231 (try 2)

juergenhoetzel:


juergenhoetzel:
Ignoring CI failure:  [Ubuntu dotnet install fails with "File has unexpected size"
](https://github.com/dotnet/core/issues/3221)

PR #243, opened 2020-08-28
Execute .NET framework binary directly when running on Windows

fbie:
Hi,

I am using this for F# development on Windows where I am living with a few restrictions and it may be useful for others, too. I have not tested this with different setups.

Kind regards, Florian


juergenhoetzel:
Thanks! I only adjusted the lisp style a bit to conform to https://github.com/bbatsov/emacs-lisp-style-guide

PR #253, opened 2020-10-28
fsharp-mode: do not explicitly depend on eglot

kiennq:
Currently `fsharp-mode` takes a hard dependency on `Eglot`, which is not desirable when other LSP system (like `lsp-mode`) has been in used.
Since user will need to take extra step of requiring `eglot-fsharp`, I think we should remove the dependency in header and instead let user install `eglot` and requiring `eglot-fsharp` by themselves

### DIFF
--- a/Cask
+++ b/Cask
@@ -5,5 +5,6 @@
 (files "*.el")
 ;; FIXME: Use multiple packages: https://github.com/melpa/melpa#example-multiple-packages-in-one-repository ?
 (development
- (depends-on "buttercup"))
+ (depends-on "buttercup")
+ (depends-on "eglot"))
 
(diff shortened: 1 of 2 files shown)


juergenhoetzel:
Good point.

The [melpa fsharp-mode](https://melpa.org/#/fsharp-mode) package should also be changed (split to `fsharp-mode` and  `eglot-fsharp`) to reflect these change. Otherwise the eglot dependency is still installed

kiennq:
@juergenhoetzel Since I'm not package owner, I doubt I can just easily create a new MELPA package for `eglot-fsharp`, can you help with that?

juergenhoetzel:
> @juergenhoetzel Since I'm not package owner, I doubt I can just easily create a new MELPA package for `eglot-fsharp`, can you help with that?

Of course.I will do. The installation notes sould be updated.

juergenhoetzel:
Merged via #263  (doc update): Thanks!

PR #210, opened 2019-07-20
Move to LSP (WIP)

juergenhoetzel:
`fsautocomplete` [protocol change to LSP](https://github.com/fsharp/FsAutoComplete/issues/410l)

TODOS:

- [x] Remove own `flycheck/company` backends
- [x] Use [Cask](https://github.com/cask/cask) instead of  non-portable `GNU  Makefile` to automate the package development cycle; development, dependencies, testing, building, packaging 
- [x] Port tests to `LSP`
- [x] Update `README.md`

PR #255, opened 2020-12-07
Fix interference with company-mode

niontrix:
Company-mode has deprecated the variables company-auto-complete and
company-auto-complete-chars in favor of company-auto-commit and
company-auto-commit-chars for the same purposes.
The references to the old variable names are kept als deprecated aliases.
Creating local variables with the deprecated variable names seems to cause
issues when activating company-mode which complains with "Don't know how to make
a localized variable an alias".
This seems to prevent reliable auto-completion.

### DIFF
--- a/fsharp-mode.el
+++ b/fsharp-mode.el
@@ -201,8 +201,8 @@
           compile-command
           syntax-propertize-function
           company-backends
-          company-auto-complete
-          company-auto-complete-chars
+          company-auto-commit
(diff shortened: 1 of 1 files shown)


juergenhoetzel:
Thank you for spotting this: These variables are a leftover from #210 (Move to LSP (WIP)).

fsharp-mode doesn't directly depend on `company-mode` anymore. I removed the variables in #256.

PR #257, opened 2020-12-19
Next round of Indentation Work

Gastove:
Hello!

Apologies this took so long; I lost steam pretty badly when the unit test framework changed. Since then, I've been able to come back to both F# *and* learn buttercup (which is honestly lovely!), and now I'm back at it. 

I don't think this PR should close anything; this work isn't complete. But, it's good progress -- a working,  non-zero test suite and _some_ improvements to how indentation works. Going to try and keep rolling with this. I'll make small updates to this PR if I find other things that seem appropriate to pull in, but otherwise I plan to mostly open new PRs for steadily more specific issues as I go.

Thanks very much!

### DIFF
--- a/fsharp-mode-structure.el
+++ b/fsharp-mode-structure.el
@@ -31,7 +31,7 @@
 ;; currently override.
 ;;
 ;; SMIE configs by m00nlight Wang <dot.wangyushi@gmail.com>, 2015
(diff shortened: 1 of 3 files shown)


Gastove:
Having opened the PR, I can now see it is already more than big enough. I wont add more to this, beyond requests made here in review :+1:

PR #258, opened 2021-01-02
Prepare split package

juergenhoetzel:
Refs #253

`eglot-fsharp` depends on `fsharp-mode`


juergenhoetzel:
26.3 failure doesn't seem to be related to this change: Need to investigate :fearful:

juergenhoetzel:
> 26.3 failure doesn't seem to be related to this change: Need to investigate

`eglot` upstream: [Flex completion style is not available on Emacs < 27
](https://github.com/joaotavora/eglot/pull/582)

PR #260, opened 2021-01-15
Fixed for split eglot-fsharp package

jcs090218:
This patch fixed most issues from https://github.com/melpa/melpa/pull/7336.

### DIFF
--- a/eglot-fsharp.el
+++ b/eglot-fsharp.el
@@ -46,8 +46,8 @@
   :group 'eglot-fsharp
   :risky t
   :type '(choice
-      (const :tag "Latest release" latest)
-      (string :tag "Version string")))
+          (const :tag "Latest release" latest)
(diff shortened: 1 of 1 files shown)


juergenhoetzel:
thanks for stepping in :+1: 

I relaxed the Emacs version dependency to 26.3 (minimum required version for eglot).

The remaining `package-lint` error `eglot-fsharp.el:157:9: error: You should depend on (emacs "27.1") if you need jsonrpc-error` seems to be a false positive. The required `jsonrpc` library is installed as transitive dependency when using Emacs 26.3.

Dependency-Tree of 26.3:
```
fsharp-mode 1.9.15
    s 1.3.1    [1.12.0 installed]
    eglot (any)    [1.7 installed]
        jsonrpc 1.0.14
```

PR #266, opened 2021-01-26
Replace dash functions and fix byte-compile warnings

syohex:
This commit https://github.com/fsharp/emacs-fsharp-mode/commit/93b3e0036121eec298fe26f1267dd61944612b7a removed dash.el dependency but some dash functions still used. This PR replaces dash functions/macros with standard library's functions/macros.


juergenhoetzel:
Nice idiomatic elisp :+1: Thanks a lot!

PR #268, opened 2021-01-31
Generic face for custom operators that contain pipe char sequence.

3Rafal:
Currently, custom operators that contain pipe char sequence are colored partially.
Let's say we have operator `|>>` (monadic map in FSharpPlus and FParsec libs).
Part that contains pipe gets colored (`|>`), part that remains (`>`) stays default.
This PR sets all custom operators with pipe sequence to default.

### DIFF
--- a/fsharp-mode-font.el
+++ b/fsharp-mode-font.el
@@ -171,6 +171,17 @@ with initial value INITVALUE and optional DOCSTRING."
   "<|\\{1,3\\}\\||\\{1,3\\}>"
   "Match the full range of pipe operators -- |>, ||>, |||>, etc.")
 
+(def-fsharp-compiled-var fsharp-custom-operator-with-pipe-regexp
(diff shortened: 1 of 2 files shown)


3Rafal:
@juergenhoetzel ,
I've added tests for operator faces. Unfortunately, when running from buttercup they fail for `|>` and `<|` cases.
Running 
```lisp
(with-highlighted "<<| |>> |> ||> |||> <| <|| <||| <|> <<|!"
  (str-face "|>"))
```
returns `fsharp-ui-operator-face` correctly though...

3Rafal:
@juergenhoetzel , I have fixed tests and they're work fine now. would you mind reviewing my PR?

juergenhoetzel:
Thank you for fixing this special-case and writing the tests :+1: 

LSP tests fail again but is unrelated to this PR  (due to `fsautocomplete` update)  :disappointed:

PR #270, opened 2021-03-31
fix: `//`-comments filled incorrectly (`//` not acting as prefix)

Kazark:
The problem is described [here](https://github.com/syl20bnr/spacemacs/issues/11326). I haven't been writing much F# lately but I found that `emacs-scala-mode` has the same problem and this fix works well for it, so I thought I would submit it here too.

### DIFF
--- a/fsharp-mode.el
+++ b/fsharp-mode.el
@@ -218,6 +218,7 @@ If set to t, the buffer will always be saved, silently."
         comment-end              ""
         comment-column           40
         comment-start-skip       "///* *"
+        adaptive-fill-regexp     "[ \t]*\\(//+[ \t]*\\)*"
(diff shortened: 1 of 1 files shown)


Kazark:
I updated the PR to not use a hook based on comments I got on my corresponding Scala PR, and testing that furthered my understanding of how `setq` interacts with major modes.

juergenhoetzel:
Thanks :+1: 

I have only made one small adjustment: https://github.com/fsharp/emacs-fsharp-mode/commit/93b1fbc31d73286a18640a36fc2be87d1736e0f2 (Make `adaptive-fill-regexp` buffer-local )

Kazark:
@juergenhoetzel Oh, yes, I see that would be necessary. Thanks!

PR #284, opened 2021-12-28
Use cl-lib function instead of cl.el's one

syohex:
I got the following error when launching REPL.

```
inferior-fsharp-mode: Symbol’s function definition is void: copy-seq
```

cl.el is deprecated since Emacs 27 and newer Emacs doesn't load by default. This file already loads `cl-lib` so it should use cl-lib functions.


juergenhoetzel:
Thanks!

PR #287, opened 2022-01-25
Switch to using npkg for server install

amcguier:
The latest release of FsAutoComplete breaks this package for me. 

This is an approach to resolving the breakage and attempting to preserve backwards compatibility with (many) previous versions of FsAutoComplete. 

Fundamentally, rather than using the .NetCore.zip which is no longer produced by the fsautocomplete release process, I've changed the server install process to grab the nupkg artifacts. This change should be compatible with fsautocomplete v0.46.1 onwards. 

If needed, I could extend the compatibility window by parsing the version and falling back to the zip file for older fsautocomplete settings. 

It's also worth noting, that while I didn't make this change, there's no longer an exe zip produced by the build for fsautocomplate for the full framework, so while my change should be compatible with previous releases it'll be broken for anything modern (as will the current release branch)

### DIFF
--- a/eglot-fsharp.el
+++ b/eglot-fsharp.el
@@ -60,12 +60,18 @@
   :type '(choice (const :tag "Use .Net Core" net-core)
                  (const :tag "Use .Net Framework" net-framework)))
 
+(defvar eglot-fsharp-server-root
+  (file-truename (concat eglot-fsharp-server-install-dir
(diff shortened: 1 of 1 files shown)


amcguier:
Closing this pull request and rethinking. 

fsharpautocomplete is published as a dotnet tool now. I'm going to rework this PR so .net core uses the local tool install rather than  manually installing/extracting

PR #296, opened 2022-04-23
`fsharp-mode-project-root`: Don't traverse filesystem twice

juergenhoetzel:
Use improved regex.


juergenhoetzel:
CI failures are unrelated and caused be recent `fsautocomplete.exe` update. Need to investigate in a different Issue/PR.

PR #298, opened 2022-04-28
tiny fix: Adds match! as a valid keyword

MMagueta:
I noticed `match!` was not being understood as a valid keyword, so I just added it.

My steps:
- nix-shell -p cask emacs python38 git which gnumake --pure
- make
- make run-fsharp-mode
- Testes highlighting on the scratch buffer after activating the mode

Note: Thanks for the great work on this repository so far

@juergenhoetzel


juergenhoetzel:
Thanks you! :+1:

PR #299, opened 2022-04-30
Replace Cask with Eask

jcs090218:
Hi, I have recently created a tool named [Eask](https://github.com/emacs-eask/eask). This is designed to replace Cask with as minimal effort as possible. The DSL syntaxes are very similar, and most commands are the same, which should be easy to adapt from Cask to Eask.

Eask also uses the same concept with Eldev. It's written only in Elisp, and the project metadata file (`Eldev` or `Eask`) is just an elisp file being loaded before the command executions.

The major difference is the `Eask` uses [yargs](https://www.npmjs.com/package/yargs) and NodeJS runtime to parse command-line arguments. Unlike `Cask` or `Eldev`, use `batch` and `bash/sh`.

#### In `.github/workflow/test.yml`

* Use `jcs090218/setup-emacs` instead of `purcell/setup-emacs`
  - Add retry capability to prevent failure on Emacs' installation
  - Support windows
* Add setup-node (`16`)
* Remove `setup-cask` (replace with `setup-eask`)
* Add test for Emacs `28.1`
* Merge jobs `ubuntu + macos` with `windows`

#### In `Makefile`

* Replace Emacs command-line arguments `emacs --batch ...` with just `eask <command>`
* Add `make checkdoc` and `make lint` but left unused (save it for later)
* Add `make build` to test packaging and installing.

#### Others

* Remove `Cask` (replace with `Eask`)

---

This is totally optional, the major goal is to cutdown/lower the entire CI complexity.


Makefile
```diff
@@ -1,40 +1,47 @@
...
+
 VERSION = $(shell git describe --tags --abbrev=0 | sed 's/^v//')
+
 PKG = fsharp-mode
 
-elpa-$(EMACS):
-	$(CASK) install
-	$(CASK) update
-	touch $@
+ci: build compile test
```
jcs090218, line 13: Here is the CI entry, now we only have three tasks.

* `build` (test `packaging` and `installing`)
* `compile` - byte-compile source files (files are defined in `Eask`-file)
* test (left unchanged)

jcs090218:
It seems like Cask is ignoring this issue. Do you think the issue should be ignored here? 😕 

```
Loading seq...
../../../../test/eglot-tests.el:Warning: Unknown LSP interface FooObject
../../../../test/eglot-tests.el:Warning: Unknown LSP interface FooObject
../../../../test/eglot-tests.el:Warning: Unknown LSP interface FooObject
../../../../test/eglot-tests.el:Warning: Unknown LSP interface FooObject
../../../../test/eglot-tests.el:Warning: Unknown LSP interface FooObject
../../../../test/eglot-tests.el:Warning: Unknown LSP interface FooObject
dotnet exited with status 1
```

PR #313, opened 2022-06-29
Add missing format specifier

syohex:


juergenhoetzel:
Nice catch :+1: Thank you!

PR #314, opened 2022-06-30
Fix byte-compile warnings

syohex:
- specify ':type' attribute for defcustom variables
- escape quotes in docstring

```
eglot-fsharp.el:54:12: Warning: defcustom for ‘eglot-fsharp-server-verbose’
    fails to specify type

fsharp-mode-structure.el:268:2: Warning: docstring has wrong usage of
    unescaped single quotes (use \= or different quoting)
```


juergenhoetzel:
Thank you!

PR #319, opened 2022-08-25
Add Additional FSAutoComplete Parameters

amcguier:
[FsAutoComplate](https://github.com/fsharp/FsAutoComplete) readme specifies additional settings provided to init that allow linting, and auto expansion of records, interfaces, etc to be supplied to eglot. 

This PR adds those parameters. I'm tempted to make this customizable via a variable, although I'm torn if it should be a single customizable value with the entire propertylist or individual values per config setting.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,9 +14,9 @@ jobs:
         os: [ubuntu-latest, macos-latest]
         dotnet: [6.0.200]
         emacs_version:
-          - 27.1
           - 27.2
           - 28.1
+          - 28.2
           - snapshot
     steps:
     - uses: purcell/setup-emacs@master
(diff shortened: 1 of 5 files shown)


amcguier:
This PR should be ready to review now. 

I opted to leave the FsAutoComplete argument as a single customizable variable. This enables several additional parameters which are already supported by `eglot` and changes the default parameters to match the recommendations for all parameters to match [FsAutoComplete ](https://github.com/fsharp/FsAutoComplete) defaults. 

I have also updated the test matrix to replace 27.1 with 28.2  since 27.1 seems to have been removed from the major package managers.

Finally I made a slight tweak to one of the integration test to reduce occasional flakiness in the test I suspect to be due to concurrency (i.e. the diagnostic message arrives before the listener is subscribed which causes the test to fail intermittently).

amcguier:
Closing in favor of #327 off of branch

PR #321, opened 2022-10-25
Update console app parameters

amcguier:
The latest version of fsautocomplete removed the `--background-service` parameter which makes eglot crash on startup. This swaps to the updated parameter for lsp mode.

`--background-service` has been deprecated, switch to experimental adaptive lsp service

change the `-v` parameter to `--verbose` to align with the documentation

### DIFF
--- a/eglot-fsharp.el
+++ b/eglot-fsharp.el
@@ -126,8 +126,8 @@ Ensure FsAutoComplete is installed (when called INTERACTIVE)."
   (when (file-exists-p (eglot-fsharp--path-to-server))
     (cons 'eglot-fsautocomplete (cons (eglot-fsharp--path-to-server)
                                       (if eglot-fsharp-server-verbose
-			                  `("--background-service-enabled" "-v")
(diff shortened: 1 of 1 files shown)


juergenhoetzel (commented):
Thank you. This also enables

`--adaptive-lsp-server-enabled                                Enable LSP Server based on FSharp.Data.Adaptive. Should be more stable, but is experimental.`

is this intended?

amcguier:
At least for me (OSX 12.6) the server crashes on boot without the adaptive flag and it runs with that flag provided. I could make that switchable with a variable if that's preferable.

juergenhoetzel:
> At least for me (OSX 12.6) the server crashes on boot without the adaptive flag and it runs with that flag provided. I could make that switchable with a variable if that's preferable.

I confirm the crash on GNU/Linux. I just added a `eglot-fsharp-server-args` to to simplify the adjustment in the future.

Thanks a lot for your contribution! :rocket:

amcguier:
Glad to help!

PR #323, opened 2022-11-08
Fix installing fsautocomplete process

syohex:
- Use cache for avoiding getting latest version
- Fix version checking
- Don't always install fsautocomplete

### DIFF
--- a/eglot-fsharp.el
+++ b/eglot-fsharp.el
@@ -64,10 +64,11 @@
 
 (defun eglot-fsharp--latest-version ()
   "Return latest fsautocomplete.exe version."
-  (let* ((json (with-temp-buffer (url-insert-file-contents "https://azuresearch-usnc.nuget.org/query?q=fsautocomplete&prerelease=false&packageType=DotnetTool")
-			         (json-parse-buffer)))
-         (versions (gethash "versions" (aref (gethash "data" json) 0))))
-    (gethash "version" (aref versions (1- (length versions))))))
+  (or eglot-fsharp--latest-version
+      (let* ((json (with-temp-buffer (url-insert-file-contents "https://azuresearch-usnc.nuget.org/query?q=fsautocomplete&prerelease=false&packageType=DotnetTool")
+			             (json-parse-buffer)))
+             (versions (gethash "versions" (aref (gethash "data" json) 0))))
+        (setq eglot-fsharp--latest-version (gethash "version" (aref versions (1- (length versions))))))))
 
 (defun eglot-fsharp--installed-version ()
   "Return version string of fsautocomplete."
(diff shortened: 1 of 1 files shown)


eglot-fsharp.el
```diff
@@ -64,10 +64,11 @@
...
   "Return latest fsautocomplete.exe version."
-  (let* ((json (with-temp-buffer (url-insert-file-contents "https://azuresearch-usnc.nuget.org/query?q=fsautocomplete&prerelease=false&packageType=DotnetTool")
-			         (json-parse-buffer)))
-         (versions (gethash "versions" (aref (gethash "data" json) 0))))
-    (gethash "version" (aref versions (1- (length versions))))))
+  (or eglot-fsharp--latest-version
+      (let* ((json (with-temp-buffer (url-insert-file-contents "https://azuresearch-usnc.nuget.org/query?q=fsautocomplete&prerelease=false&packageType=DotnetTool")
+			             (json-parse-buffer)))
+             (versions (gethash "versions" (aref (gethash "data" json) 0))))
+        (setq eglot-fsharp--latest-version (gethash "version" (aref versions (1- (length versions))))))))
```
syohex, line 71: Use cache.

https://github.com/fsharp/emacs-fsharp-mode/blob/962f2157ed229d18e4f010ff5630a3ce4b4bdc67/eglot-fsharp.el#L63

The cache variable is declared but it is never used.

eglot-fsharp.el
```diff
@@ -77,10 +78,10 @@
 (defun eglot-fsharp-current-version-p (version)
   "Return t if the installation is not outdated."
   (when (file-exists-p (eglot-fsharp--path-to-server))
-    (if (eq version 'latest)
+    (if (eq eglot-fsharp-server-version 'latest)
```
syohex, line 81: `version` argument is string type, so it is never `'latest`.

eglot-fsharp.el
```diff
@@ -77,10 +78,10 @@
...
 	(equal (eglot-fsharp--latest-version)
 	       (eglot-fsharp--installed-version))
-      (equal eglot-fsharp-server-version (eglot-fsharp--installed-version)))))
+      (equal version (eglot-fsharp--installed-version)))))
```
syohex, line 84: `eglot-fsharp-server-version` may be symbol. So `version` should be used here.

eglot-fsharp.el
```diff
@@ -96,12 +97,12 @@
 					   nil "tool" "uninstall"
 					   "fsautocomplete" "--tool-path"
 					   default-directory))
-                (error  "'dotnet tool uninstall fsautocomplete --tool-path %s' failed" default-directory))))
-          (unless (zerop (call-process "dotnet" nil `(nil ,stderr-file) nil
-				       "tool" "install" "fsautocomplete"
-				       "--tool-path" default-directory "--version"
-				       version))
-            (error "'dotnet tool install fsautocomplete --tool-path %s --version %s' failed" default-directory  version)))
```
syohex, line 104: This code block is always evaluated even if desired version is installed. I suppose if this expression should be evaluated if `(eglot-fsharp-current-version-p version)` is `nil`

eglot-fsharp.el
juergenhoetzel: The purpose of `(eglot-fsharp--latest-version)` is to ALWAYS return the exact version string, so we can compare the version string of the locally installed version string and prevent downloading the nuget artefact.

Maybe I should document this better.

syohex: @juergenhoetzel Should I remove cache variable ? Comparison is already implemented. Sorry I didn't catch what you mean.

https://github.com/fsharp/emacs-fsharp-mode/blob/f8873096d24cac7b30419854347220f37345e780/eglot-fsharp.el#L77-L83

PR #332, opened 2023-03-24
Updating the configuration object to match Ionide configuration

### LINKED ISSUE #331: fsac >= 0.59.3: No typecheck results
### Description

`eglot`  `flycheck/xref` fails for `fsac`>= 0.59.3

### Repro steps

Please provide the steps required to reproduce the problem

```bash
eldev -dtT test 
```

### Expected behavior

Integreation tests succeed

### Actual behavior

integrations checks timeout (waiting for typecheck results).

### Workaround

Pin to older version succeds:

```bash
eldev -S '(setq eglot-fsharp-server-version "0.59.2")' -dtT test 
```

amcguier:
This PR should fix #331. 

It also updates the configuration parameters to match the settings used in Ionide which enables some extra code analysis settings (warnings in the configuration). 

This also ensures that parameters get passed correctly to the `workspace/didChangeConfiguration` as well, which was omitted in the previous settings PR.

### DIFF
--- a/eglot-fsharp.el
+++ b/eglot-fsharp.el
@@ -56,23 +56,66 @@
   :type '(repeat string))
 
 (defcustom eglot-fsharp-fsautocomplete-args '(
-    :automaticWorkspaceInit t
-    :keywordsAutocomplete t
-    :externalAutocomplete nil
-    :linter t
-    :unionCaseStubGeneration t
-    :recordStubGeneration t
-    :interfaceStubGeneration t
-    :interfaceStubGenerationObjectIdentifier "this"
-    :unusedOpensAnalyzer t
-    :unusedDeclarationsAnalyzer t
-    :useSdkScripts t
-    :simplifyNameAnalyzer nil
-    :resolveNamespaces t
-    :enableReferenceCodeLens t)
-    "Arguments for the fsautocomplete initialization."
-    :group 'eglot-fsharp
-    :risky t
+                                              :automaticWorkspaceInit t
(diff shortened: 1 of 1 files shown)


amcguier:
Any chance someone could rerun the tests, looks like their were temporary timeouts during the package restoration unrelated to my changes.

juergenhoetzel:
Nice :+1:  

I have only removed the `eglot-fsharp--set-workspace-args` function in the amended commit. `eglot` already uses the initialization options by calling the 
```
(cl-defmethod eglot-initialization-options ((_server eglot-fsautocomplete)) 
...
```

amcguier:
Whoops, I left that in, I cherry picked that initial commit in from a branch I was working on. I have an experimental branch that needs to set that so you can run FsAutoComplete at a solution level rather than on each project individually.

juergenhoetzel:
BTW: The actual error in the original initialization option was the invalid `fsharp`  key at the top level of the `json` object:

```diff
-  `(:fSharp ,eglot-fsharp-fsautocomplete-args))
+  eglot-fsharp-fsautocomplete-args)
```

so I guess `fsautcomplete` never actually used the init options.

amcguier:
Yeah, I realize that the workspace/DidConfigurationChange call requires the `:fSharp` node to use the options and the intialize call expects the parameters not to be set there. I'll probably add that back in when I PR the other branch since you might change configurations (I'm working on support for workspace peek/load custom calls)

PR #334, opened 2023-06-20
Enable nxml-mode for fsproj files

augustfengd:
The `nxml-mode` is built-in and will enable editing the fsproj files like they're xml files.

### DIFF
--- a/fsharp-mode.el
+++ b/fsharp-mode.el
@@ -114,6 +114,7 @@
 
 ;;;###autoload
 (add-to-list 'auto-mode-alist '("\\.fs[iylx]?\\'" . fsharp-mode))
+(add-to-list 'auto-mode-alist '("\\.fsproj\\'" . nxml-mode))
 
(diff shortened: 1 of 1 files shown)


baronfel:
Consider trying to associate the MSBuild [Xml Schema Document](https://github.com/dotnet/msbuild/blob/main/src/MSBuild/Microsoft.Build.xsd) with this tooling for better completions!

juergenhoetzel:
Thank you :+1:

juergenhoetzel:
> Consider trying to associate the MSBuild [Xml Schema Document](https://github.com/dotnet/msbuild/blob/main/src/MSBuild/Microsoft.Build.xsd) with this tooling for better completions!

Good idea.  

Emacs needs a RELAX NG Compact Syntax (RNC) schema. 

The `Microsoft.Build.xsd` schema would need to be converted to `RNG`.

PR #345, opened 2024-12-22
Added defcustom server path

deadblackclover:
I think it would be nice if it was possible to set the path to `fsautocomplete` if it is self-installed. In my case Emacs installed in Windows 10 defines the path `~/` as `C:/Users/<User>/AppData/Roaming`, but I would like to use `fsautocomplete` installed by myself.

So I made the path `~/.dotnet/tools/` as defcustom to be able to install my own path.

### DIFF
--- a/eglot-fsharp.el
+++ b/eglot-fsharp.el
@@ -36,6 +36,11 @@
   :link '(url-link "https://github.com/fsharp/FsAutoComplete")
   :group 'eglot)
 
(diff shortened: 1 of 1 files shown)


deadblackclover:
@juergenhoetzel Could you pls review my PR

bbatsov:
The proposal seems reasonable to me.

